PR #24, opened 2023-02-05
Fix LangChain name inconsistancies

aRealCatTBH:
Longchain => LangChain

completely went over my head, my bad


dartpain:
Went over my head multiple times! So thank you so much, join our Discord, we need contributors

dartpain (approved):

PR #39, opened 2023-02-07
Add cost estimate feature

jaredbradley243:
Calculates number of tokens/user cost and requires user permission to proceed.

User permission bypass is built-in to allow for non-human users.


JorgeMag96:
This looks good, however, rather than having the same code in two different places, you could add it on it's own file and import it on both `scripts/ingest_rst.py` and `scripts/ingest_rst_sphinx.py`.

dartpain (approved):

PR #76, opened 2023-02-13
feat: compatible with markdown

mefengl:
It would be great if it could support more plain-text-like formats such as markdown, etc. I'm not sure how this repository plans to accomplish this, but I can share my solution for markdown.


dartpain (approved):

dartpain:
Great Job! thanks a lot!

PR #85, opened 2023-02-14
fix: restore index back when continue process

mefengl:
I don't know why there needs to be a `store.index = None`, but I got an error when continuing the process, so I restored the index before continuing;

But that's only fixed the index issue.

Another problem is there is no error catch for the `store.add_texts([i.page_content], metadatas=[i.metadata])` in the `except` block, so if 10 seconds is not enough (maybe set it to 60?), then the second error will be thrown out, and this time, no `try/except` for this one


dartpain:
Oh nice, thank you, definitely

PR #89, opened 2023-02-15
Redux, Clean up, Custom Hooks, Taylor Working 🚀

TaylorS15:
Tons of changes. 
All existing state is handled by Redux in src/store.ts.
useMediaQuery custom hook exported from src/hooks.ts.
Imgs folder moved to src
General cleanup of src directory.


ajaythapliyal:
The PR in its current state hasnt been able to run ESlint. Please move the below mentioned config after [this line](https://github.com/arc53/DocsGPT/blob/5883ce2685fba929e80be065847f885c4a7f978f/frontend/.eslintrc.cjs#L21)

`'prettier/prettier': [
    'error',
    {
      endOfLine: 'auto',
    },
  ]`

frontend/src/store.ts
```diff
@@ -0,0 +1,48 @@
...
+interface State {
+  isApiKeyModalOpen: boolean;
+  apiKey: string;
+  isMenuOpen: boolean;
+}
+
+const initialState: State = {
+  isApiKeyModalOpen: false,
+  apiKey: '',
+  isMenuOpen: false,
```
ajaythapliyal, line 12: IMO we should be only keeping that part of state in redux which is supposed to be shared across multiple components. I dont quite see how other components are going to need to know if menu is open or not. 

If i fail to see some usecase here, please LMK

frontend/src/store.ts
```diff
@@ -0,0 +1,48 @@
...
+
+interface State {
+  isApiKeyModalOpen: boolean;
+  apiKey: string;
+  isMenuOpen: boolean;
+}
+
+const initialState: State = {
+  isApiKeyModalOpen: false,
+  apiKey: '',
```
ajaythapliyal, line 11: These two data `isApiKeyModalOpen` and `apiKey`  see tightly correlated. 
If API key is not present => show modal.
If API key present => dont show the modal.

 If this is the case then `isApiKeyModalOpen` is derived data and we shouldnt be storing that in redux. We should be writing a selector to derive it from `apiKey`
 
` isApiKeyModalOpen = ()=> !!apiKey`

frontend/src/store.ts
```diff
@@ -0,0 +1,48 @@
...
+    },
+    setApiKey: (state, action: PayloadAction<string>) => {
+      state.apiKey = action.payload;
+      console.log('setApiKey', action.payload);
+    },
+    toggleIsMenuOpen: (state) => {
+      state.isMenuOpen = !state.isMenuOpen;
+    },
+  },
+});
```
ajaythapliyal, line 31: I see a single slice for entire app at the moment, which is fine for now. As we move ahead and things get more complex, we can split our redux store.

frontend/src/store.ts
TaylorS15: Both Navigation.tsx and APIKeyModal.tsx need access to isApiKeyModalOpen

apiKey might be able to be moved into the modal component if it is not going to be needed by Conversation.tsx when sending chats.

isMenuOpen is currently being used by both About.tsx and Conversation.tsx, and could be used by any other About-like pages in the future.

frontend/src/store.ts
TaylorS15: Dont understand this. Currently on page load apiKey is going to just be a blank string and we can choose to make isApiKeyModalOpen true after testing so the user is greeted with it when opening the page.

In the future when authentication is implemented this will probably change

frontend/src/store.ts
TaylorS15: Yeah just started learning redux today and just wanted to get something working. Will likely make multiple slices later

frontend/src/App.tsx
```diff
@@ -1,54 +1,17 @@
...
-        isApiModalOpen={isApiModalOpen}
-        setIsApiModalOpen={setIsApiModalOpen}
-      />
-      <Navigation
-        isMobile={isMobile}
-        isMenuOpen={isMenuOpen}
-        setIsMenuOpen={setIsMenuOpen}
-        setIsApiModalOpen={setIsApiModalOpen}
-      />
+    <div className="relative flex flex-col transition-all md:flex-row">
```
ajaythapliyal, line 9: Hey @TaylorS15  can you please explain what this line and styling is for?

frontend/src/components/APIKeyModal.tsx
```diff
@@ -1,34 +1,37 @@
...
-  isApiModalOpen: boolean;
-  setIsApiModalOpen: React.Dispatch<React.SetStateAction<boolean>>;
-  apiKey: string;
-  setApiKey: React.Dispatch<React.SetStateAction<string>>;
-}) {
+  toggleApiKeyModal,
+  selectIsApiKeyModalOpen,
+} from '../store';
+
+export default function APIKeyModal({}) {
```
ajaythapliyal, line 9: if the component isnt accepting any props, you could just write it as

`export default function APIKeyModal()`

unlike whats committed, above psuedocode doesnt break the linter.

ajaythapliyal:
I see there is some difference when it comes to mobile and desktop menu. 
=> the menu for mobile is displayed on the top unlike desktop which is at the bottom. It also pushes the conversation UI below the menu in mobile.

is the above difference part of the requirement? @tenokami 
image


image

ajaythapliyal:
This is what i think is happening here. All about desktop here...
1] `<Navigation>` is `position : fixed` item. Whenever we minimize or maximize the menu we change its width. Because the `<Navigation>`  is a `fixed` item we are also having to adjust the width of `<Conversation>` or `<About>`. This makes it important for all the toplevel components to know whether menu is open or not. This in turn pushes that state to redux.

I feel we can do few things to simplify this.

How about if we use flexbox? we wont have to adjust the width of each component. 
```
<div class="flex">
   <Navigation/>
   <Conversation/>
 </div>
```

in the above snippet, if we just change the width of `<Navigation>`, the conversation component will automatically adjust without having any need to know about whether menu is open or not. This will in turn not require us to push menu state in redux. Another benefit is reduced conditional statements(for classes) in `<Conversations>` / `<About>`

to handle open and closed menu, `<Navigation> ` component can keep the `isMenuOpen` state locally and return either JSX of open menu or JSX of closed menu.

frontend/src/components/APIKeyModal.tsx
TaylorS15: Just forgot to remove curly braces when removing props

tenokami:
> i
@ajaythapliyal 
here is a template for mobile sidebar.
Regarding the conversation, it should be under the menu if the menu is expanded. And the height of the conversation should be the same as the height of the menu and with scroll to navigate through the chat

image

dartpain:
Its basically a sidebar, so it slides over the chat itself from the the side

PR #115, opened 2023-02-22
Storing API Key and recent source docs locally

TaylorS15:


ajaythapliyal:
Hey Taylor, although the local storage thing works but we see a flash modal when we open the app subsequent times.
flash

frontend/src/preferences/preferenceApi.ts
```diff
@@ -0,0 +1,67 @@
...
+  }
+}
+
+export async function setLocalRecentDocs(doc: Doc): Promise<void> {
+  try {
+    localStorage.setItem('DocsGPTRecentDocs', JSON.stringify(doc));
+  } catch (error) {
+    console.log(error);
+  }
+}
```
ajaythapliyal, line 67: Any particular reasons why these local storage functions are async? AFAIK local storage web api are sync in nature. Doesnt seem like any async thing happening here.

TaylorS15: yeah I thuogth they maybe should be synchronous but wasnt sure. will change

ajaythapliyal:
Right now the way localstorage and redux are distinct and we are using them together on APIKeyModal mount. This might be the reason why the [flash](https://github.com/arc53/DocsGPT/pull/115#issuecomment-1441145769) shows up. 

There should be a better way to sync localstorage and redux together. I dont have a very concrete solution in my mind right now but looking into middlewares would be helpful. We only should be talking to redux in our component layer, rest all should be handled in other layers.

https://stackoverflow.com/questions/35305661/where-to-write-to-localstorage-in-a-redux-app
https://stackoverflow.com/questions/68421040/local-storage-using-redux-toolkit
https://github.com/rt2zz/redux-persist
https://blog.logrocket.com/persist-state-redux-persist-redux-toolkit-react/

dartpain (approved):

PR #117, opened 2023-02-23
Fix the servor 500 error and show error message to client

unkwnownGriot:
## refactor the form submitting function to show an alert when an exception is raised from the backend


dartpain (approved):

dartpain:
Great Job! thank you!

PR #120, opened 2023-02-24
Nick's feedback List : 1, 2, 4, 6

ajaythapliyal:
1] Reducing font size and maybe increasing line spacing will improve readability. Use figma as a reference

image
image

2] I'd put this lower to be more in line with the user's eye level. Use figma as a reference
image

4]Since it lacks the chat button from figma, there is no way to come back from the About page
image

6]Making the input field larger can make it easier to spot and there is no text under it like in figma
image


tenokami:
Awesome work. Thank you very much! @ajaythapliyal 

2) moving it even lower would be closer to design
4) chat button should be higlighted when we are on a chat page (like in figma)

ajaythapliyal:
> Awesome work. Thank you very much! @ajaythapliyal
> 
> 2. moving it even lower would be closer to design
> 3. chat button should be higlighted when we are on a chat page (like in figma)

@tenokami  pushed the suggested changes. LMK how it looks.

dartpain:
Perfect, I showed it to @tenokami he said it looks good!

dartpain (approved):

tenokami:
That's right

PR #130, opened 2023-02-25
Api linking

dartpain:


frontend/src/conversation/conversationSlice.ts
```diff
@@ -14,7 +14,12 @@ export const fetchAnswer = createAsyncThunk<
   { state: RootState }
 >('fetchAnswer', async ({ question }, { getState }) => {
   const state = getState();
-  const answer = await fetchAnswerApi(question, state.preference.apiKey);
+
+  const answer = await fetchAnswerApi(
+    question,
+    state.preference.apiKey,
+    state.preference.selectedDocs,
```
ajaythapliyal, line 21: `state.preference.selectedDocs!` to avoid compiler error

frontend/src/preferences/preferenceApi.ts
```diff
@@ -46,4 +46,20 @@ export function setLocalApiKey(key: string): void {
...
+    doc.language + '/' + namePath + '/' + doc.version + '/' + doc.model;
+  fetch('https://docsgpt.arc53.com/api/docs_check', {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/json',
+    },
+    body: JSON.stringify({
+      docs: docPath,
+    }),
+  }).then((response) => response.json());
```
ajaythapliyal, line 64: i feel like the failed HTTP request arent raising rejected promise properly. i checked with random api key on my local.

ajaythapliyal:
```
let namePath = selectedDocs.name;
  if (selectedDocs.language === namePath) {
    namePath = '.project';
  }

  const docPath =
    selectedDocs.language +
    '/' +
    namePath +
    '/' +
    selectedDocs.version +
    '/' +
    selectedDocs.model;

  return fetch('https://docsgpt.arc53.com/api/answer', {
    method: 'POST',
    headers: {
      'Content-Type': 'application/json',
    },
    body: JSON.stringify({
      question: question,
      api_key: apiKey,
      embeddings_key: apiKey,
      history: localStorage.getItem('chatHistory'),
      active_docs: docPath,
    }),
  })
    .then((response) => {
      if (response.ok) {
        return response.json();
      } else {
        Promise.reject(response);
      }
    })
    .then((data) => {
      const result = data.answer;
      return { answer: result, query: question, result };
    });
```

Alternate way which rejects the promise when HTTP req doesnt succeed.
fetch promises are successful even http request fails, so we gotta manually check `response.ok`

ajaythapliyal: BTW i am not confident if this covers all conditions. Please test and make necessary modifications.

frontend/src/conversation/conversationApi.ts
```diff
@@ -1,28 +1,47 @@
...
+    fetch('https://docsgpt.arc53.com/api/answer', {
+      method: 'POST',
+      headers: {
+        'Content-Type': 'application/json',
+      },
+      body: JSON.stringify({
+        question: question,
+        api_key: apiKey,
+        embeddings_key: apiKey,
+        history: localStorage.getItem('chatHistory'),
```
ajaythapliyal, line 33: is this line going to be replaced with something else?

frontend/src/conversation/conversationApi.ts
```diff
@@ -1,28 +1,47 @@
...
+      }),
+    })
+      .then((response) => response.json())
+      .then((data) => {
+        const result = data.answer;
+        resolve({ answer: result, query: question, result });
+      })
+      .catch((error) => {
+        reject();
+      });
```
ajaythapliyal, line 44: i feel like the failed HTTP request arent raising rejected promise properly. i checked with random api key on my local.


```
let namePath = selectedDocs.name;
  if (selectedDocs.language === namePath) {
    namePath = '.project';
  }

  const docPath =
    selectedDocs.language +
    '/' +
    namePath +
    '/' +
    selectedDocs.version +
    '/' +
    selectedDocs.model;

  return fetch('https://docsgpt.arc53.com/api/answer', {
    method: 'POST',
    headers: {
      'Content-Type': 'application/json',
    },
    body: JSON.stringify({
      question: question,
      api_key: apiKey,
      embeddings_key: apiKey,
      history: localStorage.getItem('chatHistory'),
      active_docs: docPath,
    }),
  })
    .then((response) => {
      if (response.ok) {
        return response.json();
      } else {
        Promise.reject(response);
      }
    })
    .then((data) => {
      const result = data.answer;
      return { answer: result, query: question, result };
    });
```

Alternate way which rejects the promise when HTTP req doesnt succeed.
fetch promises are successful even http request fails, so we gotta manually check `response.ok`

BTW i am not confident if this covers all conditions. Please test and make necessary modifications.

frontend/src/App.tsx
```diff
@@ -19,7 +19,7 @@ export default function App() {
       />
       <div
         className={`transition-all duration-200 ${
-          navState === 'ACTIVE' ? 'ml-0 md:ml-72 lg:ml-60' : ' ml-0 md:ml-16'
```
ajaythapliyal, line 22: are these whitespaces changes intentional?

frontend/package.json
```diff
@@ -48,6 +48,6 @@
     "prettier-plugin-tailwindcss": "^0.2.2",
     "tailwindcss": "^3.2.4",
     "typescript": "^4.9.5",
-    "vite": "^4.1.0"
+    "vite": "^4.1.4"
```
ajaythapliyal, line 51: is this intentional too?

frontend/src/preferences/preferenceApi.ts
ajaythapliyal, line 64: is this sequence of operation correct?

user selects a document > save the doc to LS > make API request to check if Doc is available

I am not sure about the details here but I feel because theres no error handling in http req there can be a case where LS and backend might be in an inconsistent state.

example :
Do we still want to keep it saved in LS when http req has failed?

If the current implementation does the job we can keep it the way it is now and think about it few days later 😉

frontend/src/conversation/conversationApi.ts
dartpain: In future - yes

frontend/src/App.tsx
dartpain: seems to be @TaylorS15 changes

frontend/package.json
dartpain: Again @TaylorS15 changes

frontend/src/preferences/preferenceApi.ts
dartpain: Agreed need to handle errors here and loading states

PR #136, opened 2023-02-27
chat-small-fixes

dartpain:


frontend/src/Avatar.tsx
```diff
@@ -5,5 +5,5 @@ export default function Avatar({
   avatar: string;
   size?: 'SMALL' | 'MEDIUM' | 'LARGE';
 }) {
-  return <div>{avatar}</div>;
+  return <div className={'mt-4 text-2xl'}>{avatar}</div>;
```
ajaythapliyal, line 8: I think Avatar should not be putting margin top. Here Margin top is a need by its parent. So i think correct way here is make the styling customizable by passing the class name from parent.

example : 
https://github.com/arc53/DocsGPT/blob/b5cb3f994bfab561852b84487a1b33a82108bd92/frontend/src/conversation/ConversationBubble.tsx#L11

pabik (approved):

PR #154, opened 2023-03-05
Refactors singular Message model to Query model & adds api request to post feedbacks

ajaythapliyal:
1] Changes the conversation model from single Message type to Query which would contain prompt and its response.
2] Adds an API request to send feeback.
3] Adds env files


dartpain (approved):

dartpain:
Tested it work, lgtm

PR #217, opened 2023-04-22
Added celery to requirements

Zillibub:
The celery library was missing from the requirement.txt, added it

### DIFF
--- a/application/requirements.txt
+++ b/application/requirements.txt
@@ -11,6 +11,7 @@ blinker==1.5
 blobfile==2.0.1
 boto3==1.26.102
 botocore==1.29.102
+celery==5.2.7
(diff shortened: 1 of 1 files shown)


dartpain:
Thank you so much for your PR, but wierdly enough we do need to change the way container build too to make it work, as currently celery, gunicorn and tiktoker are being installed in Dockerfile

Zillibub:
Yep, found it. 
Why do you install it in Dockerfile?

dartpain:
Because it used via cli. And i use a trick for installing tiktoken to increase compatability on different platforms, which impacts the way dependencies are installed. I think i can change dockerfile to install all requirements from the requirements.txt but I have to test it a bit

PR #222, opened 2023-04-27
Verify CPU AVX compatibility

darth-pika-hu:
Integrate a function to verify CPU AVX compatibility, preventing users from attempting to use the software with an incompatible CPU. This is crucial as MongoDB 5.0 and later versions necessitate a CPU with AVX instruction set support.

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -1,4 +1,16 @@
 #!/bin/bash
+
+check_cpu_avx_support() {
+  if grep -q -E "avx" /proc/cpuinfo; then
+    echo "AVX support found."
+  else
+    echo "Error: Your CPU does not support the AVX instruction set, which is required by Mongo 5+."
+    exit 1
+  fi
+}
+
+# Check if the CPU supports AVX
+check_cpu_avx_support
 cd "$(dirname "$0")" || exit
 
 # Create the required directories on the host machine if they don't exist


dartpain:
Another good commit, but witl this work on mac?

darth-pika-hu:
Please help me to try.
——
Bryan Hu
***@***.***

dartpain:
Hmm try changing check_cpu_avx_support to this
```
check_cpu_avx_support() {
    if [[ $(uname -s) == "Darwin" ]]; then
        sysctl -a | grep -E 'hw.optional.avx1_0: 1' > /dev/null
    else
        grep -q -E "avx" /proc/cpuinfo > /dev/null
    fi

    if [ $? -eq 0 ]; then
        echo "AVX support found."
    else
        echo "Error: Your CPU does not support the AVX instruction set, which is required by Mongo 5+."
        exit 1
    fi
}
```

Thank you!

darth-pika-hu:
How about this one:

> 
> #!/bin/sh
> 
> check_cpu_avx_support() {
>   case "$(uname -s)" in
>     Linux)
>       if grep -q -e avx /proc/cpuinfo; then
>         return 0
>       else
>         return 1
>       fi
>       ;;
>     Darwin) # macOS
>       if sysctl -a | grep -q -e machdep.cpu.features.*AVX; then
>         return 0
>       else
>         return 1
>       fi
>       ;;
>     *)
>       echo "Unsupported OS"
>       return 1
>       ;;
>   esac
> }
> 
> if check_cpu_avx_support; then
>   echo "AVX support found."
> else
>   echo “Error: Your CPU does not support the AVX instruction set, which is required by Mongo 5+."
>   exit 1
> fi

——
Bryan Hu
***@***.***

PR #228, opened 2023-05-07
Update app.py

riccardofresi:
handle history on openai_chat


dartpain:
Thank you so much for this PR but i have to reject it as i implemented history but differently.
I named you as a coauthor for it🫡

PR #250, opened 2023-05-26
Update .env-template

tardigrde:
Rename env var name to OPENAI_API_KEY.
This was needed for me to run the docker-compose (prod) setup.

Few questions:
- what about API_KEY? is it used anywhere else than inside the docker containers?
- what about EMBEDDINGS_KEY? is it needed? also see that it is used in containes (only?)

### DIFF
--- a/.env-template
+++ b/.env-template
@@ -1,2 +1,2 @@
-API_KEY=<LLM api key (for example, open ai key)>
+OPENAI_API_KEY=<LLM api key (for example, open ai key)>
 EMBEDDINGS_KEY=<LLM embeddings api key (for example, open ai key)>


tardigrde:
Well I realized, that in the https://github.com/arc53/DocsGPT#quickstart it says:
`Create .env file in your root directory and set your OPENAI_API_KEY with your openai api key`

For me this was confusing as I expected that i'd just have to rename .env-template to .env, set my key and I'm good to go. I did not realize that in the quickstart a new .env has to be setup, for which we do not have a template. Is that right?

I set the PR as draft, I guess there is no need to merge it now. What would be nice, is a more consoldiated way of treating env vars. I can help you with that, if you have an idea how you would like to have it.

dartpain (approved):
Sorry for taking so long, thank you so much must have missed that

PR #265, opened 2023-06-15
Improve dev setup docs

tardigrde:
I wanted to play around with the code, but it was a bit struggle for me, so I'd like to contribute my findings.

As wrote in discord, I had a problem on a debian fresh install with `pip install -r application/requirements.txt`, I get error:
```
subprocess.CalledProcessError: Command '['./build_bundled.sh', '0.1.97']' returned non-zero exit status 127.
      [end of output]
  
  note: This error originates from a subprocess, and is likely not a problem with pip.
  ERROR: Failed building wheel for sentencepiece
  Running setup.py clean for sentencepiece
```
When I remove the sentencepiece specific version (==0.1.97), pip figures it out and can install all packages without an issue. Is there a reason you have hardcoded version numbers? I did not change this tho. But I'm interested why do I have this [problem](https://sebhastian.com/python-error-legacy-install-failure/#:~:text=The%20error%3A%20legacy%2Dinstall%2D,moved%20to%20a%20different%20name.)

tiktoken and celery were also missing for running `wsgi.py`. I added them to requirements, but without a hardcoded version. Should I add some static versions?

npx was needed for the husky hook. and I figured out Node v14 is not enough for that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,7 +110,7 @@ pip install -r requirements.txt
 6. Start worker with `celery -A app.celery worker -l INFO`
 
 ### Start frontend 
-Make sure you have Node version 14+
+Make sure you have Node version 16+
 
 1. Navigate to `/frontend` folder
 2. Install dependencies
--- a/README.md
+++ b/README.md
@@ -73,30 +73,50 @@ To stop just run Ctrl + C
 
 ## Development environments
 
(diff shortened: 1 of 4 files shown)


dartpain:
Fixed container build.

Thank you so much!

in terms of sentencepiece i also pushed a change

dartpain (approved):

tardigrde:
Thanks Alex!

PR #269, opened 2023-06-17
Azure implementation for the streaming output

larinam:
1. Support Azure for streaming output
2. Shell script for running the whole system through docker-compose depending on the availability of Azure environment variables.


dartpain (approved):

dartpain:
Good commit, Thank you!

PR #272, opened 2023-06-23
Fixed bugs, changed backend port

idan-h:
Hey, there was a bug when you sent more than a few messages, the content would just get too long for a GET request, and the ui just got stuck.
I changed the stream method to POST.

There was another bug I fixed when tried to use the documentation embedding indexing.

Also, changed the port to a less used one as I had port conflicts.


dartpain (approved):
Thank you so much!

PR #275, opened 2023-07-01
feat: use generic Modal with render props

Sukhacoder02:
This PR extracts out the Modal functionality into a generic Modal component onto which actual views are rendered through render props.

---
[Both of the existing modal components](https://www.figma.com/file/OXLtrl1EAy885to6S69554/DocsGPT?node-id=0%3A1&t=Xqi6tyl2G3mghdiP-0) have few things in common:
1]They both overlay over existing page. Both have similar CSS design(rounded corners, shadows).
2]Both have some similar functions as well, both will go away when "save" is clicked. As for now one of them have cancellation feature but as we add more modals, this too can be a common functionality.

These things should be moved to a generic component. This modal will accept :
1]save click handler as a prop
2]isCancellable boolean prop to decide whether to enable the cancellable feature or not.
3]The actually view to render as [render prop](https://www.patterns.dev/posts/render-props-pattern/)

### DIFF
--- a/frontend/src/Modal/index.tsx
+++ b/frontend/src/Modal/index.tsx
@@ -7,6 +7,7 @@ interface ModalProps {
   render: () => JSX.Element;
   modalState: string;
   isError: boolean;
+  errorMessage?: string;
 }
 const Modal = (props: ModalProps) => {
   return (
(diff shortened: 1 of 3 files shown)


dartpain:
Thank you so much!

dartpain (approved):

Sukhacoder02:
Thenks a ton! @dartpain

PR #295, opened 2023-08-05
Fix min_tokens logic for grouping documents

larinam:
documents with (length >= min_tokens) should not be grouped into one document for indexing.


dartpain (approved):
Thank you!

Issue #276, opened 2023-07-08
openai.error.APIConnectionError when asking a question to a document


RikSchoonbeek:
In the development environment, as set up using the instructions [here](https://github.com/arc53/DocsGPT#development-environments), when asking a question to a document an openai.error.APIConnectionError occurs. To be complete, it's a POST request to `http://localhost:7091/api/answer` in which the error occurs. See Traceback below.

The cause seems to be the clashing of the variable name OPENAI_API_BASE used in DocsGPT and the variable name used by the openai Python package. Another project had the same issue, [see here](https://github.com/Significant-Gravitas/Auto-GPT/pull/648).

This can be solved simply by renaming OPENAI_API_BASE as used in the DocsGPT code to another name, like OPENAI_AZURE_API_BASE.

I already made the changes locally and will make a PR, as other people seem to run into the same issue, see:
https://discord.com/channels/1070046503302877216/1070605825112035399/1118266133305376869
https://discord.com/channels/1070046503302877216/1070605825112035399/1118296369883512864

Error Traceback:
```
Traceback (most recent call last):
  File "/home/rik/.virtualenvs/DocsGPT2/lib/python3.11/site-packages/openai/api_requestor.py", line 596, in request_raw
    result = _thread_context.session.request(
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rik/.virtualenvs/DocsGPT2/lib/python3.11/site-packages/requests/sessions.py", line 575, in request
    prep = self.prepare_request(req)
           ^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/rik/.virtualenvs/DocsGPT2/lib/python3.11/site-packages/requests/sessions.py", line 486, in prepare_request
    p.prepare(
  File "/home/rik/.virtualenvs/DocsGPT2/lib/python3.11/site-packages/requests/models.py", line 368, in prepare
    self.prepare_url(url, params)
  File "/home/rik/.virtualenvs/DocsGPT2/lib/python3.11/site-packages/requests/models.py", line 439, in prepare_url
    raise MissingSchema(
requests.exceptions.MissingSchema: Invalid URL '/engines/text-embedding-ada-002/embeddings': No scheme supplied. Perhaps you meant https:///engines/text-embedding-ada-002/embeddings?

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/rik/projects/DocsGPT2/application/app.py", line 326, in api_answer
[...]

PR #296, opened 2023-08-05
Revert "Changed environment variable names OPENAI_API_BASE and OPENAI…

larinam:
…_API_VERSION to AZURE_OPENAI_API_BASE and AZURE_OPENAI_API_VERSION"

This reverts commit ce8b29e9d0fbc6c5b33194446d56e9ad945ccde4.

Please see the documentation for langchain.embeddings.openai.OpenAIEmbeddings to get the point of the environment varibles naming. These names should not be changed.


larinam:
Hi @dartpain , 
Could you review this one? If additional explanation is required, I can provide it.
The thing with naming is that in the langchain package, it is expected to have particular names for the environment variables. When the change which I suggest to revert here was introduced, it led to the broken Azure support in DocsGPT. The revert of the commit will recover functionality which was in place before the change introduced by the commit, which is suggested to be reverted in this pull request.
Thank you for your time reviewing these changes and maintaining this project!

dartpain:
Thank you for you explanation, it seems to be working fine for me on the dev environment and quickstart one 

#276 I just want to be careful not to break it again.

dartpain (approved):

PR #299, opened 2023-08-13
fixes for requirements.txt

larinam:
1. remove unnecessary dependencies to speed up the build and reduce the download volume.
2. upgrade a couple of dependencies.


dartpain (approved):
great! Thank you!

PR #300, opened 2023-08-14
Introduce tests with pytest

larinam:
Necessary packages/modules modification.
Pytest introduction.


dartpain (approved):
Thank you so much, very good PR!

PR #306, opened 2023-08-18
Adapt documentation to existing tests.

larinam:
This PR is only about documentation change, please review.

### DIFF
--- a/.github/workflows/pytest.yml
+++ b/.github/workflows/pytest.yml
@@ -1,13 +1,14 @@
-name: Run python tests with pytest
+name: Run Python tests with pytest
 
 on: [push, pull_request]
+
+permissions:
+  pull-requests: write
 
 jobs:
   pytest_and_coverage:
     name: Run tests and count coverage
-    permissions:
-      contents: read
-      pull-requests: write
+
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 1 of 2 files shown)


larinam:
Hi @dartpain ,

To have coverage action adding a comment to the pull request, the action should have permission to write to the pull request. It is configured inside the action:
Screenshot 2023-08-18 at 17 49 52
But when I check GITHUB_TOKEN permissions, I see that it is not modified. 
Screenshot 2023-08-18 at 17 49 11
Most likely permissions related to the GitHub token used in the repository (secrets.GITHUB_TOKEN) are limited and don't allow to write to pull requests.
Could you check permissions related to the GITHUB_TOKEN connected to the repo and grant permissions to write to pull requests to allow action to add comments to the pull request with the coverage statistics?

larinam:
OK, according to the section from the article: https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token it is impossible to get the write permissions when merging from the public forked repository. So,

larinam:
I'll prepare the pull request with the partial revert of coverage PR.

dartpain:
Yeah, I was trying to understand that for a bit, whats the right way to set it up?

larinam:
I suppose this particular action will be functional only for PRs within one repository and doesn't work for public forks. So we can't use it here. I'll look for another solution. Maybe something like SonarCloud or similar. I'll create an issue about it or create another pull request if find a proper solution.
But this PR is only about the documentation and is still relevant and looks safe to be merged :)

dartpain:
Thank you!

PR #312, opened 2023-09-03
PR coverage with barecheck github app

larinam:
This approach uses the GitHub application https://github.com/apps/barecheck to acquire pull-request write permissions.

To make it work, please follow the next steps:
1. Install the GitHub application https://github.com/apps/barecheck for the repository DocsGPT.
2. Write down the token that the application produces.
3. Copy the token provided on the authorization confirmation page and [add it to your build environment](https://docs.github.com/en/actions/learn-github-actions/variables#creating-configuration-variables-for-a-repository) as BARECHECK_GITHUB_APP_TOKEN. I recommend adding it **as a secret** rather than as a variable.
4. After that feel free to rerun the new GitHub action workflow.
5. Merge the PR :)

Happy to clarify or answer any questions.

The build will fail if the coverage decreases after the commit. Also, a similar comment will be added to each Pull Request:
Screenshot 2023-09-03 at 21 34 23

### DIFF
--- a/.github/workflows/pr_coverage.yml
+++ b/.github/workflows/pr_coverage.yml
@@ -1,0 +1,67 @@
+name: Code Coverage
+
+on: [pull_request]
+
+jobs:
+  base_branch_cov:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          ref: ${{ github.base_ref }}
+      - name: Set up Python 3.11
+        uses: actions/setup-python@v4
+        with:
+          python-version: "3.11"
+      - name: Install dependencies
+        run: |
+          python -m pip install --upgrade pip
+          pip install pytest pytest-cov
+          cd application
+          if [ -f requirements.txt ]; then pip install -r requirements.txt; fi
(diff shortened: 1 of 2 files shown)


larinam:
So, did you put the CODECOV_TOKEN as plain text so that it works for forks?

dartpain:
As i understand it only needs token for main branch, but for pull requests it works. There where few more set up steps, but i guess They handled auth in a similar way to vercel. 
Looks like its working here:
 #316

dartpain:
Quick update, its probably because my fork was from pre-coverage version

dartpain:
Seems to be working fine here #317

larinam:
But what is you add BARECHECK_GITHUB_APP_TOKEN not as a secret, but rather as a variable? This thing should work then, right?

dartpain:
I tried both, and they mentioned that variables are not used in the workflows that are triggered by a pull request from a fork. 
Can you please try creating another PR, but make sure its synced first.

https://community.codecov.com/t/how-to-get-coverage-for-a-pr-from-a-foreign-fork-with-github-actions/1437

PR #318, opened 2023-09-04
add simple test to make a PR to check CodeCov

larinam:


dartpain:
Try to sync your fork and then create a pr, this way it will be able to compare commits

PR #325, opened 2023-09-07
remove old static resources from the Flask application, update the ro…

larinam:
remove old static resources from the Flask application, update the routing in app.py


dartpain:
Thank you! We definetly needed to get rid of that

dartpain (approved):

PR #363, opened 2023-10-01
feat(ui): add scroll to bottom button

### LINKED ISSUE #170: UI Update. Added arrow to go back to the end of conversation
Added an arrow to go back to the end of conversation when the convesation is long enough. The arrow appears when the user scrolls away from the last message
arrow

On movile the arrow should be transparent when it overlaps text. Use ChatGPT as a referece

hirenchauhan2:
This will show the scroll to bottom button when user scrolls to top from the last message.

Closes #170

### DIFF
--- a/frontend/src/assets/arrow-down.svg
+++ b/frontend/src/assets/arrow-down.svg
@@ -1,0 +1,3 @@
+<svg xmlns="http://www.w3.org/2000/svg" height="24" viewBox="0 -960 960 960" width="24">
+  <path d="M440-800v487L216-537l-56 57 320 320 320-320-56-57-224 224v-487h-80Z" />
+</svg>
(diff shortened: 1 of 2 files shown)


hirenchauhan2:
For developing this part I had to use the mock data from the app as I was not able to run the back-end app due to some issues in my python installation in WSL. I'm not super comfortable in fixing python side of issues. I also commented the code which shows the `Select Source Documentation` modal.

hirenchauhan2:
In the preview, the `Select Source Documentation` modal does not go away due to CORS error from the API.

dartpain:
Good commit,
There is one small thing you need to add before ill merge the PR, 

Basically if user is at the bottom of the screen, the arrow should dissapear.
I see that you tried to implement it in code but I guess you have to tweak few things.

Also even if you can't launch backend just try it with error messages too

Btw looks good on mobile too!

dartpain:
Just added [endMessageRef.current] to deps, seems to fix the issue

dartpain (approved):

dartpain:
@holopin-bot @hirenchauhan2

Issue #382, opened 2023-10-03
Empty Spaces Passed in the Input Field


ArnabBCA:
The Input Field does not handle Empty Spaces
![Emp](https://github.com/arc53/DocsGPT/assets/113178195/06cd443a-bd04-4624-840d-a5a422995073)

ArnabBCA:
Can i Work on this issue?

ArnabBCA:
> Please do, thank you!
> 

I have submitted a PR. please check

PR #384, opened 2023-10-03
Fixed Empty Spaces Passed in the Input Field

ArnabBCA:
Issue ID #382

The Input Field now handles Empty Spaces

### DIFF
--- a/frontend/src/conversation/Conversation.tsx
+++ b/frontend/src/conversation/Conversation.tsx
@@ -61,6 +61,8 @@ export default function Conversation() {
   };
 
   const handleQuestion = (question: string) => {
+    question = question.trim();
+    if (question === '') return;
     dispatch(addQuery({ prompt: question }));
     dispatch(fetchAnswer({ question }));
   };


dartpain:
I think we still send the empty spaces if a user submits it as a request, maybe we should rather stop the send button from working if its empty. Somewhere around handleQuestion when it gets triggered in Conversation.tsx.

thank you!

ArnabBCA:
> I think we still send the empty spaces if a user submits it as a request (it will basically send no empty spaces otherwise), maybe we should rather stop the send button from working if its empty. Somewhere around handleQuestion when it gets triggered in Conversation.tsx.
> 
> thank you!

Thank you for your feedback I will look into the code and make necessary changes.

ArnabBCA:
> I think we still send the empty spaces if a user submits it as a request (it will basically send no empty spaces otherwise), maybe we should rather stop the send button from working if its empty. Somewhere around handleQuestion when it gets triggered in Conversation.tsx.
> 
> thank you!

I have moved the **_trim_** logic in the **handleQuestion** function as  you  told, now even if the user press enter or click button the handleQuestion will first check whether the question is empty or not then it will proceed else it will return . it will also not send any request to the backend also

dartpain (approved):

dartpain:
@holopin-bot @ArnabBCA Thank you!

PR #409, opened 2023-10-04
Huggingface quantisation

Archit-Kohli:
- **Documentation Update**

- **Why was this update required?**: Difficult to load model with such high parameters so needs to be loaded via bitsnbytes

- **Other information**: Added link to hugging face's official bits-n-bytes notebook for reference


dartpain:
Unfortunately users will need to change some code in application/llm/huggingface.py
What you can do is maybe create a differnt class or a q parameter that can be passed to it called q to set quantisation if explicit by users. Will only need to add few lines of code here

dartpain:
Unfortunately users will need to change some code in application/llm/huggingface.py
What you can do is maybe create a differnt class or a q parameter that can be passed to it called q to set quantisation if explicit by users. Will only need to add few lines of code here

Archit-Kohli:
Updated in a new pull request #425 you can close this request

Issue #77, opened 2023-02-13
Encapsulate the decision of what view to render(mobile or desktop) within the component. 


ajaythapliyal:
As per the current code [we are maintaining the view data as a state](https://github.com/arc53/DocsGPT/blob/e88ff885fe4391ffede61a601edacfc0d8a4b5c0/frontend/src/App.tsx#L10) in parent component and [passing it down to children using props](https://github.com/arc53/DocsGPT/blob/e88ff885fe4391ffede61a601edacfc0d8a4b5c0/frontend/src/App.tsx#L45) which can get out of hand pretty quickly as soon as we have many component that have seperate views for mobile and desktop.

Instead we could let each component make their own independent decision whether or when to switch to mobile view or desktop view by encapsulating the decision within the component. 

Since this is a common logic shared across multiple components, this can be moved into a [custom hook](https://fireship.io/snippets/use-media-query-hook/) which then can be used by multiple components.

PR #420, opened 2023-10-04
useMediaQuery

jerempy:
Feature. 
The issue is here: https://github.com/arc53/DocsGPT/issues/77

There had been some work done on this since that issue was made, but the ask hadn't been done yet - custom hook - so some props still needed to be passed even though they could be not needed.

This PR removes the dependencies on passing props to the `Navigation`
It also adds a hook to use like `const {isMobile} = useMediaQuery();`
And then can run any checks anywhere for rendering based on if mobile or not.

Also added `isDesktop` in case there is a need for future display settings for desktop size and bigger.

refactored the navState to use this new hook and renamed to navOpen to be more explicit on what it is.

- **Other information**:
I wrote a test, but there is no testing library in here, so i just commented it out and left it for the future.
Everything looks good and as expected.

https://github.com/arc53/DocsGPT/assets/95110820/879652c1-85e9-445e-855c-9a5c2ef498ad


dartpain:
This is the error message

src/hooks/index.test.ts(1,1): error TS1208: 'index.test.ts' cannot be compiled under '--isolatedModules' because it is considered a global script file. Add an import, export, or an empty 'export {}' statement to make it a module.
--
<span>Error: Command "npm run build" exited with 2</span>



Thank you for contributing!

dartpain (approved):
Great PR, Thank you!

dartpain:
@holopin-bot @jerempy

Issue #426, opened 2023-10-05
🐛 Bug Report: Overflowing text when user inputs a long string without spaces


ArnabBCA:
### 📜 Description

When user inputs a long string without spaces it overflows

### 👟 Reproduction steps

Enter a long string without spaces 
![lonh string](https://github.com/arc53/DocsGPT/assets/113178195/924ce81c-6767-4192-b228-b7d31bb83fd3)

### 👍 Expected behavior

It should break works according to parent container size.

### 👎 Actual Behavior with Screenshots

It should not overflow when user inputs a long string without spaces

### 💻 Operating system

Windows

### What browsers are you seeing the problem on?

Chrome

### 🤖 What development environment are you experiencing this bug on?

Local dev server

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

Yes, I am willing to submit a PR!

### 🧑‍⚖️ Code of Conduct

- [X] I agree to follow this project's Code of Conduct

staticGuru:
@ArnabBCA 
It's an excellent approach to contribute.

dartpain:
Closed in #427

PR #427, opened 2023-10-05
Fixed Overflowing Text messages with long strings without spaces

ArnabBCA:
- **What kind of change does this PR introduce?** (Bug fix)

#426 
- **Other information**:
- By giving a style of **"break-all"** to add line breaks whenever necessary, without trying to preserve whole words.

Before
lonh string
After
Fixed long


dartpain (approved):

ArnabBCA:
> Thank you!

Bro I think you have closed the wrong Issue, my Issue was different it's still open

PR #476, opened 2023-10-07
Fix missing documentation for using Llama_cpp

### LINKED ISSUE #466: 🚀 Feature: run with CPU using llama_cpp guide
### 🔖 Feature description

Add a feature where we run in CPU using llama_cpp

This feature need because we can run this Doc GPT in cpu without using any GPU 

It will aim achieve the model to run on CPU 

### 🔄️ Additional Information

I tried to run this model in CPU it overload the cpu would get crash so I use above model and converted into gguf model the i run the model in cpu 

None

mnickrogers:
- **What kind of change does this PR introduce?** Docs update.

- **Why was this change needed?** Explains how to use Llama_cpp as the selected LLM, resolves #466.

- **Other information**: N/A


mnickrogers:
Sure thing. From the looks of it, they only need to download that model if they haven't run setup.sh and are just editing the env config, right?

dartpain:
Yep, exactly!

mnickrogers:
Alright, should be up-to-date. Let me know if it makes sense.

dartpain (approved):

dartpain:
@holopin-bot @mnickrogers Thank you!

PR #479, opened 2023-10-07
Create Pinup

Jamwodwox:

- **Other information**:


Jamwodwox:
Hello

larinam:
Thank you for contributing! Please look into the feedback below.

1. This is the wrong place for GitHub Actions.
2. Why does it duplicate many steps we have in other actions?
3. What is the reason for having this change at all?

dartpain:
We already have workflows for similar functionality,
Thank you!

PR #499, opened 2023-10-07
Improve and optimize the Dockerfile for the application

Chamindu36:
- Improve and optimize the Dockerfile for the application

- Docker file is not optimized enough and lacks readbility

- **Other information**:

### DIFF
--- a/application/Dockerfile
+++ b/application/Dockerfile
@@ -1,23 +1,40 @@
+# Stage 1: Builder stage for Tiktoken
 FROM python:3.10-slim-bullseye as builder
 
-# Tiktoken requires Rust toolchain, so build it in a separate stage
+# Install required dependencies for building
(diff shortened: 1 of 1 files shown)


dartpain:
I think only # Stage 1: Builder stage for Tiktoken and other stage 2 comments are useful. Other comments are self explanatory. Please remove them.

Thank you so much!

larinam:
Explanation to @Chamindu36 :
```A good best practice to follow when writing comments in code is to describe the why, not the what. In general, good code should be self-explanatory in how it works, at least in isolation. Explain the reasoning behind the code, not just what the code is doing. This can help others understand the intent of the code.```
https://swimm.io/learn/code-collaboration/comments-in-code-best-practices-and-mistakes-to-avoid

PR #515, opened 2023-10-08
Fix: adjust left margin on content container

timoransky:

This very small PR fixes a bug for the wrong left margin on the content container which was not in line with the sidebar width.

Part of the content was hidden behind the sidebar on certain viewport sizes.

| Before | After |
|--------|-------|
|docsgpt arc53 com | docsgpt arc53 com (1) |


dartpain:
@holopin-bot @timoransky  Thank you!

dartpain (approved):

PR #546, opened 2023-10-10
Update .env-template to improve user friendliness

shrivastavanolo:
- **Improves User friendliness** 

- **Why was this change needed?**: As a fairly new coder even a seemingly obvious thing like -- in the .env file the env variables should be strings is not so obvious. So, I've modified the .env file to be more user friendly.

- **Other information**: Please tag this as hacktoberfest-accepted!

### DIFF
--- a/.env-template
+++ b/.env-template
@@ -1,8 +1,8 @@
-API_KEY=<LLM api key (for example, open ai key)>
-VITE_API_STREAMING=true
+API_KEY="<LLM api key (for example, open ai key)>"
+VITE_API_STREAMING="True"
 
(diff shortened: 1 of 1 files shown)


adeyinkaezra123:
I think this would end up confusing people. Inasmuch that, with the recent versions of Node etc. they are equal. With or without, it does not matter, and in some environments, quoting values will end up causing issues e.g non-unix systems (some Windows systems), 

It's generally a good practice to avoid quoting environment variable values unless it's explicitly required by the system or configuration you're working with or when you have special characters in your secret keys, might it be necessary to wrap your values within quotations _which in this case, they won't!_.

Good job though 👍

PR #598, opened 2023-10-14
refactor(frontend): simplify JSX conditional rendering

Juneezee:

JSX conditional rendering can be simplified to use the logical AND operator (`&&`) [^1] instead of ternary operator (`? :`) if we want to render something only when the conditon is true, and nothing otherwise.

[^1]: https://react.dev/learn/conditional-rendering#logical-and-operator-

- **Other information**:


dartpain (approved):

dartpain:
@holopin-bot @Juneezee

PR #601, opened 2023-10-14
Update README.md

nikcode9:

- **Other information**:


dartpain:
Thank you @nikcode9 but we do not use Angular in any way.

Issue #590, opened 2023-10-12
🐛 Bug Report: Directions of dropdown arrow misaligned.


vedant-z:
### 📜 Description

Direction of dropdown arrow should be downwards when not selected and upwards when selected. 

![image](https://github.com/arc53/DocsGPT/assets/93431609/902d5ea5-1378-479b-bc81-7906bd4d6f65)

### 👟 Reproduction steps

Go to main page. See the sidebar.

### 👍 Expected behavior

Direction of dropdown arrow should be downwards when not selected and upwards when selected.

### 👎 Actual Behavior with Screenshots

![image](https://github.com/arc53/DocsGPT/assets/93431609/902d5ea5-1378-479b-bc81-7906bd4d6f65)

### 💻 Operating system

Windows

### What browsers are you seeing the problem on?

Chrome

### 🤖 What development environment are you experiencing this bug on?

Docker

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

Yes, I am willing to submit a PR!

### 🧑‍⚖️ Code of Conduct

- [X] I agree to follow this project's Code of Conduct

vedant-z:
@dartpain Please assign me this issue.

PR #619, opened 2023-10-16
fix: Direction of dropdown arrow corrected

vedant-z:
Issue: #590

### DIFF
--- a/frontend/src/Navigation.tsx
+++ b/frontend/src/Navigation.tsx
@@ -264,7 +264,7 @@ export default function Navigation({ navOpen, setNavOpen }: NavigationProps) {
               src={UploadIcon}
               onClick={() => setUploadModalState('ACTIVE')}
             ></img>
-            {isDocsListOpen && (
+            {!isDocsListOpen && (
               <div className="absolute top-12 left-0 right-6 ml-2 mr-4 max-h-52 overflow-y-scroll bg-white shadow-lg">
                 {docs ? (
                   docs.map((doc, index) => {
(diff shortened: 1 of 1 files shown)


dartpain:
Please make sure dropdown is not open by default, thank you!

frontend/src/Navigation.tsx
```diff
@@ -264,7 +264,7 @@ export default function Navigation({ navOpen, setNavOpen }: NavigationProps) {
               src={UploadIcon}
               onClick={() => setUploadModalState('ACTIVE')}
             ></img>
-            {isDocsListOpen && (
+            {!isDocsListOpen && (
```
krabbi, line 267: name of the variable, in that case, should be `isDocsListClosed`, but to prevent negations and rename variables comfier will be to change the ternary [line 258](https://github.com/arc53/DocsGPT/pull/619/files#diff-32e7c0fb6a467d3fa7d164653fb13981bc6ba4a08cba2d8ce1b4ebdac24b2b4dL258)

vedant-z: @krabbi I don't think there is need to change the ternary as it has similar functionality to the arrow on navigation.

Please see the ternary at line [184](https://github.com/arc53/DocsGPT/blob/66f2e549ceea8d2af8c7bade657e455d617d290f/frontend/src/Navigation.tsx#L184).

Please correct me if I am wrong.

vedant-z: But as @dartpain pointed out the list is open by default so I need to fix that.

dartpain (approved):

dartpain:
Thank you @holopin-bot @vedant-z

PR #684, opened 2023-10-24
fix: virtual environment setup typo

Ayushpanditmoto:
typo fixing

Screenshot_1

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ b) On Windows
 
 ```commandline
 python -m venv venv
- venv/Scripts/activate
+source venv/Scripts/activate
 ```
 
 3. Change to the `application/` subdir by the command `cd application/` and install dependencies for the backend:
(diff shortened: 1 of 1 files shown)


larinam (changes requested):
I see the following in the documentation. Maybe the previous section requires fixing, not this, Windows one?
Screenshot 2023-10-24 at 17 23 02

Ayushpanditmoto:
@larinam  is this fine now ??

Ayushpanditmoto:
i just made change for different command for different terminal commands

larinam (changes requested):
I honestly think we don't need to extend this section of documentation. It should be basic knowledge of operating systems and Python/python environment documentation.
https://docs.python.org/3/tutorial/venv.html#creating-virtual-environments

If you want to help, please fix the problem in line 135 and that is it. And replace the link in line 129 with the one I shared above.

larinam:
Unnecessary change.

PR #712, opened 2023-10-27
Update README.md fix grammar

Archit-Kohli:
Removed article 'the' from 2 places and added state of the art to emphasize how good it is 😁

- **What kind of change does this PR introduce?** Documentation update

- **Why was this change needed?** Incorrect grammar in readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,7 +7,7 @@
 </p>
 
 <p align="left">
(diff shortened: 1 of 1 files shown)


Archit-Kohli:
@dartpain can you merge this one?

dartpain:
Thank you so much for your contribution. I suggest you check out docs/pages for a place to contribute with text.
This change is not related to grammar, its phrasing. We are trying to keep text here simple without adding more words.

PR #727, opened 2023-10-29
Add "Scroll to Top" in Readme

PrajwalGraj:
Added "Back to Top" button in the Readme.md file.

***Description***
As the readme.md is very huge and log this button will help the users to scroll to top in just one click .

Feature

As the Readme was huge and after reading the entire file ,we need to scroll the screen manually to reach the top.So this button will help to do the same in just one click

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,3 +1,4 @@
+<div id="top"></div>
 <h1 align="center">
   DocsGPT  🦖
 </h1>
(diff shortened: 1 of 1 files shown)


dartpain:
Thank you so much, appreciate your PR. 
I suggest you check out docs/pages as a place to commit to.

PR #729, opened 2023-10-30
Added comments for improved code readability in py2doc.py

THEGAMECHANGER416:
Added comments for improved code readability for other contributors

- **Updated py2doc.py to improve readability for contributors.**

- **Appropriate comments with the code help understand the code much more easily and quickly.**

- **Introduced a non-breaking change that can merged instantly**

### DIFF
--- a/scripts/parser/py2doc.py
+++ b/scripts/parser/py2doc.py
@@ -7,6 +7,7 @@ from langchain.llms import OpenAI
 from langchain.prompts import PromptTemplate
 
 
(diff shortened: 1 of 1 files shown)


dartpain:
Thank you so much for choosing to contribute to docsgpt.
Comments should describe reason behind having some of those functions.
I would suggest you add comments in places that are hard to understand instead of relatively simple functions.
Check out worker.py, I think it will benefit from description of what inputs it needs for ingest task

Issue #724, opened 2023-10-29
🚀 Feature: Use S3- bucket as the vector store


jolo-dev:
### 🔖 Feature description

The user should be able to add an S3 bucket for storing and accessing their documents.

### 🎤 Why is this feature needed ?

The documents are not just stored in the cloud but are also easier to share.
That would reduce the storage usage on your hard drive.

### ✌️ How do you aim to achieve this?

In order to store documents in an S3, you can pass a variable `S3_STORE=my-bucket-name` via the `.env` file. However, if you are running the application on your local machine, you will need to provide AWS credentials. The good news is that you can choose how to provide these credentials: https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html

When running `scripts`, the result should be uploaded to the given S3-bucket.

The store in the application should access the documents from the S3-bucket.

It could look like this

```python

from langchain.embeddings import OpenAIEmbeddings
from langchain.vectorstores import FAISS
from langchain.document_loaders import S3Loader

embeddings = OpenAIEmbeddings() 

if(os.getenv("S3_STORE")):
  loader = S3Loader(os.getenv("S3_STORE"))
  documents = loader.load()

faiss_index = FAISS.from_documents(documents, embeddings) 

# Save index to S3
faiss_index.save(os.getenv("S3_STORE") + "/faiss-index")

# Load index from S3 
faiss_index = FAISS.load(os.getenv("S3_STORE") + "/faiss-index")
```

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

jolo-dev:
@jaredbradley243 thanks for your interest.
I am currently working on that but if you want you can assign this to you 😉

jaredbradley243:
That's very kind, but If you're already working on it, keep going! 😃

jolo-dev:
@jaredbradley243 No no. Really. Let me be your reviewer then ;)

jaredbradley243:
Hahah. It's gonna take me a bit of time to work on this. I'll need to re-familiarize myself with the codebase and I won't have time to get started for a week or two, but if everyone is all right with waiting, I'll happily take it off your hands!

dartpain:
@jaredbradley243  No worries!
Ill check back in a few weeks on this

jaredbradley243:
> @jaredbradley243 Any update on this ????

[...]

PR #807, opened 2023-12-25
Update Scripts to integrate AWS S3

jaredbradley243:
This PR modified the ingest and open_ai_func files to integrate AWS S3 into DocsGPT
https://github.com/arc53/DocsGPT/issues/724

- **Other information**:

### DIFF
--- a/scripts/ingest.py
+++ b/scripts/ingest.py
@@ -16,7 +16,12 @@ from parser.py2doc import transform_to_docs
 from parser.schema.base import Document
 from parser.token_func import group_split
 
-dotenv.load_dotenv()
+import boto3
+import botocore
+from boto3.s3.transfer import TransferConfig
+
+# When override=True .env variables take precedence over environment. https://pypi.org/project/python-dotenv/
+dotenv.load_dotenv(override=True)
 
 app = typer.Typer(add_completion=False)
 
@@ -49,6 +54,8 @@ def ingest(yes: bool = typer.Option(False, "-y", "--yes", prompt=False,
            token_check: Optional[bool] = typer.Option(True, help="Whether to group small documents and split large."),
            min_tokens: Optional[int] = typer.Option(150, help="Minimum number of tokens to not group."),
            max_tokens: Optional[int] = typer.Option(2000, help="Maximum number of tokens to not split."),
+           use_s3: Optional[bool] = typer.Option(False, "--s3", help="Whether to use S3 as the document store"),
+           s3_assume_role: Optional[bool] = typer.Option(False, "--s3-assume", help="Whether to use S3 as the document store"),
            ):
     """
         Creates index from specified location or files.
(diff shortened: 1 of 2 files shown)


jaredbradley243:
## Script Functionality

1. **Local Mode (`default`)**: Processes documents from local directories specified by the user.
2. **S3 Mode (`--s3`)**: 
   - Downloads documents from an S3 bucket to a temporary local storage (`s3_temp_storage`).
   - Processes these documents.
   - Uploads the processed documents back to the S3 bucket.

## Enabling S3 Storage

To enable S3 storage, use the `--s3` flag when running the script.

1. **Environment Variables**: Set these variables in your `.env` file:
   - `S3_BUCKET`: Name of your S3 bucket.
   - `S3_DOCUMENTS_FOLDER`: Folder within the S3 bucket where your documents are stored. If left blank, all files in S3 bucket will be downloaded (except .faiss and .pkl).
   - `S3_SAVE_FOLDER`: Folder within the S3 bucket in which you would like to save the vector files. Leave blank to use the root of the bucket.

2. **Running the Script**: 
    - python ingest.py ingest --s3


## Enabling Role Assumption

If accessing an S3 bucket requires assuming an IAM role (e.g., for cross-account access), the script supports this through the `--s3-assume` flag and proper AWS configuration.

1. **Environment Variable**: 
- Add `AWS_ASSUME_ROLE_PROFILE` to your `.env` file with the name of the AWS profile for role assumption. Ex: `AWS_ASSUME_ROLE_PROFILE="dev"`

1. **AWS Configuration**:
- **Credentials File** (`~/.aws/credentials`):
  ```ini
  [default]
  aws_access_key_id = YOUR_DEFAULT_ACCESS_KEY
  aws_secret_access_key = YOUR_DEFAULT_SECRET_KEY

  [iamadmin]
  aws_access_key_id = EXAMPLEKEY123456
  aws_secret_access_key = EXAMPLESECRETKEY123456
  ```
- **Config File** (`~/.aws/config`):
  ```ini
  [default]
  region = us-west-2
  output = json

  [profile dev]
  region = us-west-2
  role_arn = arn:aws:iam::123456789012:role/YourRoleName
  source_profile = iamadmin
  ```

3. **Running the Script with Role Assumption**:

   - python your_script.py --s3 --s3-assume

This configuration allows the script to assume `YourRoleName` using the credentials from the `iamadmin` profile.

## Note
- Ensure that the IAM role (`YourRoleName`) has necessary permissions for accessing the specified S3 bucket.
- The script will create a temporary local storage (`s3_temp_storage`) for processing S3 documents, which will be cleaned up after processing.

scripts/ingest.py
```diff
@@ -16,7 +16,12 @@
 from parser.schema.base import Document
 from parser.token_func import group_split
 
-dotenv.load_dotenv()
+import boto3
+import botocore
+from boto3.s3.transfer import TransferConfig
+
+# When override=True .env variables take precedence over environment. https://pypi.org/project/python-dotenv/
+dotenv.load_dotenv(override=True)
```
larinam, line 24: Why has this changed? Default logic looks reasonable.

scripts/ingest.py
```diff
@@ -49,6 +54,8 @@ def ingest(yes: bool = typer.Option(False, "-y", "--yes", prompt=False,
            token_check: Optional[bool] = typer.Option(True, help="Whether to group small documents and split large."),
            min_tokens: Optional[int] = typer.Option(150, help="Minimum number of tokens to not group."),
            max_tokens: Optional[int] = typer.Option(2000, help="Maximum number of tokens to not split."),
+           use_s3: Optional[bool] = typer.Option(False, "--s3", help="Whether to use S3 as the document store"),
+           s3_assume_role: Optional[bool] = typer.Option(False, "--s3-assume", help="Whether to use S3 as the document store"),
```
larinam, line 58: Wrong help-message.

scripts/ingest.py
```diff
@@ -79,22 +86,105 @@ def process_one_docs(directory, folder_name):
...
-        folder_counts[folder_name] += 1
-        if folder_counts[folder_name] > 1:
-            folder_name = f"{folder_name}_{folder_counts[folder_name]}"
-        folder_names.append(folder_name)
-
-    for directory, folder_name in zip(dir, folder_names):
-        process_one_docs(directory, folder_name)
+            get_user_permission(docs, folder_name, use_s3, s3_assume_role)
+
+    if not use_s3:
```
larinam, line 93: Please review the code below and extract meaningful methods to avoid multi-screen functions.
I recommend reading a couple of Kent Beck's books.

scripts/parser/open_ai_func.py
```diff
@@ -77,8 +80,51 @@ def call_openai_api(docs, folder_name):
         c1 += 1
     store.save_local(f"outputs/{folder_name}")
 
-
-def get_user_permission(docs, folder_name):
+    if use_s3:
```
larinam, line 83: extract the method for s3 logic

larinam:
To me, the following parameters look more logical to pass over command line parameters rather than as environment variables.
S3_BUCKET
S3_SAVE_FOLDER
AWS_ASSUME_ROLE_PROFILE

larinam (changes requested):

scripts/ingest.py
jaredbradley243: Without `override=true`, the environment takes precedence over the .env variables. If a user should change their AWS credentials in the .env file, I wanted the change to take effect immediately. 😀

scripts/ingest.py
jaredbradley243: Great catch! Thank you. I'll fix that!

scripts/ingest.py
larinam: Sorry, I didn't get the justification and the use case.

jaredbradley243:
Thanks for the change requests, @larinam. I'll get to these soon!

jaredbradley243:
Working on these now.

jaredbradley243:
> To me, the following parameters look more logical to pass over command line parameters rather than as environment variables. S3_BUCKET S3_SAVE_FOLDER AWS_ASSUME_ROLE_PROFILE

I added command line parameter support as well as env variables. 😁

scripts/ingest.py
jaredbradley243: tldr, if override is not set to true, the aws credentials in .env will not update each time the script is run. 

Use case: User wants to change credentials in their env file between script runs.

scripts/ingest.py
jaredbradley243: Resolved

scripts/parser/open_ai_func.py
jaredbradley243: Resolved

scripts/ingest.py
larinam: I don't see your clear understanding of this change. I trust the default behaviour is more generally acceptable and in this case as well.
Please provide a clear case or remove this change.

scripts/parser/open_ai_func.py
```diff
@@ -93,8 +98,57 @@ def get_user_permission(docs, folder_name):
...
         print("The API was not called. No money was spent.")
+
+def upload_to_s3(s3_bucket, s3_save_folder, s3_assume_role, aws_assume_role_profile):
+    # Set defaults from environment variables if not provided via command line
+    s3_bucket = s3_bucket or os.environ.get("S3_BUCKET")
+    s3_save_folder = s3_save_folder or os.environ.get("S3_SAVE_FOLDER")
+    aws_assume_role_profile = aws_assume_role_profile or os.environ.get("AWS_ASSUME_ROLE_PROFILE")
+    
+    if not s3_bucket:
+        print("Error: S3_BUCKET environment variable is not set.")
```
larinam, line 114: Please get acquainted with the concept of Errors and Exceptions in Python.

larinam: Also the message is not completely correct in the context of the latest changes.

larinam:
In general, according to the ticket, support for S3 should be added to the application as well. In this PR the changes are only in the scripts package. Please refer to the initial ticket for more information.

larinam:
To avoid passing s3 config through several calls, please consider using https://docs.pydantic.dev/latest/concepts/pydantic_settings/
The configuration is static and doesn't change through the lifecycle of the script, so please consider using static object to manage these settings.

larinam (changes requested):
Please review the comments in the code and general comments.

Issue #809, opened 2023-12-26
🐛 Bug Report: Upload Documents Names field is editable Causing Wrong file Upload Name.


ArnabBCA:
### 📜 Description

The Name Field in the Upload Documents is Editable causing wrong file upload name.
As a result incorrect document name is stored in the backend.

### 👟 Reproduction steps

1. Go to https://docsgpt.arc53.com/
2. Click on Upload Button
3. Upload any document
4. Change the Document Name in the Name Field
5. Click on Train
6. Go to the Settings then Documents Tab
7. See the Wrong File Upload Name

### 👍 Expected behavior

The Name Field Should not allow to be editable. It should only allow actual document name.

### 👎 Actual Behavior with Screenshots

Original Document Name is MyDocument.pdf
![D1](https://github.com/arc53/DocsGPT/assets/113178195/2829b931-36b2-4cea-bbde-ee154f890746)

Writing Wrong document Name
![D2](https://github.com/arc53/DocsGPT/assets/113178195/d0bdd0eb-c427-4b9a-861d-503a5b654347)

Incorrect Document Name is Stored
![D3](https://github.com/arc53/DocsGPT/assets/113178195/93f9b6fe-52db-4a40-bee4-d5ac1b1cf2ec)

### 💻 Operating system

Windows

### What browsers are you seeing the problem on?

Chrome

### 🤖 What development environment are you experiencing this bug on?

Docker

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

Yes, I am willing to submit a PR!

### 🧑‍⚖️ Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ArnabBCA:
I want to work on this issue can you assign me.

Jessy-py:
How can I help out with this issue?

ArnabBCA:
@dartpain submited the PR pls take a look

PR #812, opened 2023-12-31
Fixed Upload File Name Field

ArnabBCA:
- **What kind of change does this PR introduce?** (Bug fix)

- #809 

- **Other information**:
- The Name Field in the Upload Documents was Editable causing wrong file upload name. As a result incorrect document name is stored in the backend.
- By Putting "**readonly**" attribute in the Input Field  solved the problem


ArnabBCA:
@dartpain pls review my PR and Happy New Year :)

dartpain:
Issue is, the filename should be editable

dartpain:
We have to make sure we use edited name in the vectors table

PR #863, opened 2024-02-29
feat: Qdrant vectorstore support

Anush008:
## Description

This PR intends to add support for Qdrant - https://qdrant.tech/ to be used as a vectorstore in DocsGPT.

You can try this out by setting the `QDRANT_LOCATION` value to `":memory:"` to use an in-memory instance for prototyping.

For using Qdrant server, run
```console
docker run -p 6333:6333 qdrant/qdrant
```
and set `QDRANT_LOCATION` to `"http://host:6333"`

### DIFF
--- a/application/core/settings.py
+++ b/application/core/settings.py
@@ -3,6 +3,7 @@ from typing import Optional
 import os
 
 from pydantic_settings import BaseSettings
+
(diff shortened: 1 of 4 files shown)


dartpain (approved):
Good PR, thank you!

PR #881, opened 2024-03-21
Fix model selection at least for openAI LLM_NAME

larinam:
There is a weird only partly working behaviour with the model_name selection. I fixed it and added another setting to configure — MODEL_NAME.

I couldn't select gpt-4 model for OpenAI. Now I can.

- **Other information**:
Pretty small fix.


dartpain (approved):
Thank you, good PR to manage different models!

Issue #887, opened 2024-03-23
🚀 Feature: Default document change to empty document


pabik:
### 🔖 Feature description

Change default document to empty (no document provided to the model). If the empty document is selected Chunks processed per query should be 0. Add current default as DocsGPT.

 

### 🎤 Why is this feature needed ?

Currently default document has DocsGPT documentation which in my opinion is confusing.

### ✌️ How do you aim to achieve this?

In the Source Doc dropdown change default to empty. While selected no document is used while chatting. 

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

dartpain:
Done, assigned.

PR #895, opened 2024-03-30
added feature #887

sarfarazsiddiquii:
This PR (ref #887 ) introduces a new feature that changes the default document selection to "Empty" and ensures that when no document is selected, chunks processed per query are set to 0. 

Currently default document has DocsGPT documentation which was confusing.

### DIFF
--- a/application/api/answer/routes.py
+++ b/application/api/answer/routes.py
@@ -207,7 +207,9 @@ def stream():
         prompt_id = data["prompt_id"]
     else:
         prompt_id = 'default'
-    if 'chunks' in data:
+    if 'selectedDocs' in data and data['selectedDocs'] is None:
+        chunks = 0
+    elif 'chunks' in data:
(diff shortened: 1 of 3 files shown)


dartpain:
Thank you so much for your PR.
There are a few adjustments that are needed to make it work:
Inside `frontend/src/conversation/conversationApi.ts` you need to make sure that
`fetchAnswerApi`, `fetchAnswerSteaming` and `searchEndpoint` can handle a new type for a `selectedDocs: Doc,`
After this slight adjustment it should all work well.

sarfarazsiddiquii:
Thank you @dartpain for your feedback. I've updated fetchAnswerApi, fetchAnswerSteaming, and searchEndpoint to correctly handle selectedDocs: Doc | null as suggested

dartpain:
Good PR, all works well, thank you!

PR #899, opened 2024-04-03
feat: added prompts section under general in settings

siiddhantt:
- dropdown for selecting prompts and editing/deleting them as per Figma design
- modals for adding and editing prompts


dartpain:
Thank you!

Issue #945, opened 2024-05-10
🐛 Bug Report: Local model is recreated for each request, causing delay and out of memory errors


starkgate:
### 📜 Description

I am using the docsgpt-7b-mistral.Q8_0.gguf model on a 1080 Ti. The model fits in the VRAM with 300MB to spare. I can query the model and it works. However, I noticed that querying the model a second time causes an out of memory error. After investigation, this is caused by DocsGPT creating a new instance of the model to answer my second query, instead of reusing the previous one or at least removing the first one before creating a new instance. Even if I had enough memory, creating a new instance causes an unnecessary delay with each query.

### 👟 Reproduction steps

I am running DocsGPT 0.9.0 locally (./setup.sh with option 2). The environment is configured to use llama-cpp with huggingface_sentence-transformers/all-mpnet-base-v2 embeddings. To reproduce the bug, monitor the VRAM with nvidia-smi then ask at least 2 questions with this environment. Notice how the VRAM usage either doubles (if you have enough VRAM) or DocsGPT runs out of memory.

### 👍 Expected behavior

The LLM instance should be cached for future queries.

### 👎 Actual Behavior with Screenshots

DocsGPT runs out of memory

### 💻 Operating system

Windows

### What browsers are you seeing the problem on?

Firefox, Chrome

### 🤖 What development environment are you experiencing this bug on?

Docker

### 🔒 Did you set the correct environment variables in the right path? List the environment variable names (not values please!)

CELERY_BROKER_URL
CELERY_RESULT_BACKEND
EMBEDDINGS_NAME
FLASK_APP
FLASK_DEBUG
LLM_NAME
VITE_API_STREAMING

### 📃 Provide any additional context for the Bug.

I have fixed this issue on my side by modifying the llama_cpp.py script to cache and reuse LLM instances. This requires the flask-caching package and some modifications to llms, retrievers and the answer API.

Here is the implementation in llama_cpp.py:
```python
singleton_llm = {
    'type': None,
    'llm': None
}

def create_llm(self, type, api_key, user_api_key, *args, **kwargs):
    llm_class = self.llms.get(type.lower())
    if not llm_class:
        raise ValueError(f"No LLM class found for type {type}")

    # do not create a new LLM (and allocate memory again) for each request for local models
    if self.singleton_llm['type'] != llm_class or self.singleton_llm['type'] != LlamaCpp:
        llm = llm_class(api_key, user_api_key, *args, **kwargs)
[...]

PR #966, opened 2024-05-28
Cache local model to reduce memory usage and delays

starkgate:

Store the LLM after creation, for future use. This is especially necessary for local LLMs. Previously the LLM was recreated for each request, which caused long delays and doubled the GPU memory consumption.

Fixes https://github.com/arc53/DocsGPT/issues/945

- **Other information**:

### DIFF
--- a/application/api/answer/routes.py
+++ b/application/api/answer/routes.py
@@ -11,7 +11,6 @@ from bson.objectid import ObjectId
 
 
 from application.core.settings import settings
-from application.llm.llm_creator import LLMCreator
 from application.retriever.retriever_creator import RetrieverCreator
 from application.error import bad_request
 from application.cache import cache
--- a/application/api/answer/routes.py
+++ b/application/api/answer/routes.py
@@ -14,6 +14,7 @@ from application.core.settings import settings
 from application.llm.llm_creator import LLMCreator
 from application.retriever.retriever_creator import RetrieverCreator
 from application.error import bad_request
+from application.cache import cache
 
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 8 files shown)


dartpain:
Sorry for a delay on this PR. Really appreciate it!
Main reason is because its a relative large conceptual change.
Personally I think there should be no LLM inference on our API container. Ideally it should be a separate one.
One kind of inference that I think is ok for the time being is embeddings, but it also causes issues with memory, I had multiple issues because of it on our production servers, so I moved embedding creation to a singleton instead of classic flask cache. 

[PR for embeddings](https://github.com/arc53/DocsGPT/pull/996)

Think both cache and singleton patterns are good, but I would prefer singleton pattern due to its more simplistic nature. Maybe because Im more used to it too.

Is there any advantage or disadvantage to either one that you see?

starkgate:
> Sorry for a delay on this PR. Really appreciate it! Main reason is because its a relative large conceptual change. Personally I think there should be no LLM inference on our API container. Ideally it should be a separate one. One kind of inference that I think is ok for the time being is embeddings, but it also causes issues with memory, I had multiple issues because of it on our production servers, so I moved embedding creation to a singleton instead of classic flask cache.
> 
> [PR for embeddings](https://github.com/arc53/DocsGPT/pull/996)
> 
> Think both cache and singleton patterns are good, but I would prefer singleton pattern due to its more simplistic nature. Maybe because Im more used to it too.
> 
> Is there any advantage or disadvantage to either one that you see?

No worries on the delay. I chose the flask cache option since it seemed easier to implement: we need the cache/singleton to be accessible throughout the API, as a sort of global variable. Flask cache made this simple, just call the cache and you have your object. I wasn't sure how to do this cleanly with a singleton, admittedly I'm also not a Python expert, I could rework the PR to use a singleton instead of cache.

dartpain:
Hey @starkgate Thank you so much for this PR, I do think we will go in favour of a singleton.
Also thank you so much for bringing it up as it brought up some other issues in our application. We definitely need to change things. If you can re-work it as a singleton - would be amazing!

starkgate:
> Hey @starkgate Thank you so much for this PR, I do think we will go in favour of a singleton. Also thank you so much for bringing it up as it brought up some other issues in our application. We definitely need to change things. If you can re-work it as a singleton - would be amazing!

I'm glad to hear it helped!

Issue #998, opened 2024-06-18
🐛 Bug Report: Language switch to Japanese or Mandarin, blackouts the Settings tab selection


ManishMadan2882:
### 📜 Description

In Settings > General
* On changing the language to Mandarin or Japanese the tab selection is lost
[Watch Demo to the bug](https://www.webmobilefirst.com/en/screencasts/2-DMoU_PTS/)

### 👟 Reproduction steps

1. Go to Settings > general
2. Switch language under the Language Selection dropdown to Japanese or Mandarin
3. Any tab can still be selected in the Settings window but the state of current tab is missed 

### 👍 Expected behavior

The language switch should smoothly translate the activeTab

### 👎 Actual Behavior with Screenshots

[Watch Demo of the issue](https://www.webmobilefirst.com/en/screencasts/2-DMoU_PTS/)

### 💻 Operating system

Linux

### What browsers are you seeing the problem on?

Chrome

### 🤖 What development environment are you experiencing this bug on?

Local dev server

### 🔒 Did you set the correct environment variables in the right path? List the environment variable names (not values please!)

Yes

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

None

### 🧑‍⚖️ Code of Conduct

- [X] I agree to follow this project's Code of Conduct

utin-francis-peter:
Hi, @dartpain. 

Kindly assign me the issue.

utin-francis-peter:
Closed the prev PR due to some conflicts in commit.  Below is an updated PR:

PR #1008, opened 2024-06-23
Fix/issue#998

utin-francis-peter:
a fix for issue#998

To do the following:
- improve UX by not rendering an already active language in the list of languages to choose from
- correctly rendering the "General" component when language option is changed to Mandarin or Japanese

- **Other information**:
closes: https://github.com/arc53/DocsGPT/issues/998


ManishMadan2882 (approved):
Perfect @utin-francis-peter , everything works as expected!

PR #1012, opened 2024-06-25
fix: input box improvements

### LINKED ISSUE #1010: fix: input box
If we use multiple lines for input in our input box
CleanShot 2024-06-25 at 10 09 17@2x

Some of the text is hidden by the curves on it
Also there is too much space between each line

siiddhantt:
- fixes #1010


dartpain (approved):
LGTM! Thank you!

PR #1013, opened 2024-06-25
fix: use singleton in llama_cpp

dartpain:

- **Other information**:


starkgate:
@dartpain Looks good, thanks! Much cleaner than my proposal, too. This will be a great improvement for many people.

pabik (approved):

Issue #1022, opened 2024-07-02
🐛 Bug Report: Code Display Overflow and Formatting Issue


shelar1423:
### 📜 Description

The code display section does not handle code formatting properly, causing the code to overflow horizontally and extending beyond its intended bounds.

### 👟 Reproduction steps

1. Insert or paste code into the code display section/query bar(mostly the code with many indentations and spaces).
2. Observe the rendering of the code.

### 👍 Expected behavior

The code should be properly formatted within the container, with line breaks and proper indentation to ensure readability. The code display section should handle overflow correctly, either by wrapping the code within the container or providing a scrollable view.

### 👎 Actual Behavior with Screenshots

![image](https://github.com/arc53/DocsGPT/assets/82649533/f4a85a49-d379-4f71-b292-4340db5b4159)

### 💻 Operating system

Windows

### What browsers are you seeing the problem on?

Chrome

### 🤖 What development environment are you experiencing this bug on?

Docker

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

None

### 🧑‍⚖️ Code of Conduct

- [X] I agree to follow this project's Code of Conduct

shelar1423:
> Working on this. please assign me the issue in the meantime

Assigned it, thank you!

moink7:
Hi, im working on this issue, please assign me!

PR #1030, opened 2024-07-03
FIX: #1022 / fixed code text overflowing issue

blackviking27:
- FIX

Fixed the overflowing of the code
- **Other information**:
Issue: [https://github.com/arc53/DocsGPT/issues/1022](https://github.com/arc53/DocsGPT/issues/1022)


siiddhantt:
hi @blackviking27 thanks for the contribution!
The whitespace handling looks much better now. However, I noticed that when code is pasted it still shows spaces for the line breaks. Could you take another look at that?
Screenshot

Issue #1016, opened 2024-06-27
Feature: Set default i18n language depending on browsers language


dartpain:
Set i18n initial language using browsers default one, make sure they are able to change it later

utin-francis-peter:
Hi, @dartpain 

I am on this issue now. 

Please confirm: if a user's default browser language is set to a language that isn't English nor our currently supported language options, default in that case should be set to English?

dartpain:
Yeep, thats correct!

PR #1032, opened 2024-07-04
FEAT: Auto Language Detection using User's Browser Default

utin-francis-peter:
feature: on first visit, user's browser language is detected and used if supported. Otherwise, English is used.

Users no longer have to set their primary language manually in the app.

- **Other information**:
closes: https://github.com/arc53/DocsGPT/issues/1016


utin-francis-peter:
> Thankyou @utin-francis-peter !

Always a pleasure, grand masti.

Issue #1036, opened 2024-07-15
🚀 Feature: Attachment source doc


laronson:
### 🔖 Feature description

Give users a method of selecting source documents that is integrated into the query writing process.  To do this, add a hotkey to the input text box which opens a selection popup and allows users to select from a list of their available source documents.  Once an option is selected, display the selection above the input text box to give users better insight into which documents they are using for their query.

https://github.com/user-attachments/assets/1d0bdd5d-67a1-40f4-b5a5-c3349995686e

### 🎤 Why is this feature needed ?

This feature provides a few benefits to the user:
- The hot key doc selection mechanism may enhance the query writing workflow for some users as it allows user focus to remain closer to the input text box without needing to shift focus to the source docs dropdown on the left side of the page.
- Highlighting the document that is being used for the current query closer to the input text box may reduce confusion for some users as to which document is selected for use with their current query.
- This new selection mechanism could be used in the future to scale the number of document included in a search when/if the project moves to allow more than one document to be used in a single query.  

### ✌️ How do you aim to achieve this?

1. Add a popup selector to the project as a common component to use for document selection outside of the source docs dropdown.
2. Add a source document pill list on top of the conversation input text box to show the user which document is selected for the query they are writing
3. Refactor the conversation input text box to use the new popup selector and pill list components.
- For a proposed hot key, I think ctrl + d could be a good selection.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

sOnU1002:
i can do it please assign tis to me

dartpain:
This issue, I really cool. We will probably do a heavy redesign in terms of sources.
Really appreciate you idea @laronson 
Once we come up with a cool design idea for it we can continue working on it.

action-apps:
@sOnU1002 @dartpain 

Thanks for the shoutout!  I would love to help get this going when you find time.  I have an open PR with some changes to help with this.  I know you have been working on change to the UI over the last few months but I would be happy to refactor my PR with those new changes if this were something that you wanted to integrate into the project.

laronson:
@sOnU1002 @dartpain

Thanks for the shoutout! I would love to help get this going when you find time. I have an open PR with some changes to help with this. I know you have been working on change to the UI over the last few months but I would be happy to refactor my PR with those new changes if this were something that you wanted to integrate into the project.

[...]

PR #1037, opened 2024-07-15
#1036 Conversation Input Document Selector With Hot Key

laronson:
- Feature Request Feature Request #1036 

- The hot key doc selection mechanism may enhance the query writing workflow for some users as it allows user focus to remain closer to the input text box without needing to shift focus to the source docs dropdown on the left side of the page.
- Highlighting the document that is being used for the current query closer to the input text box may reduce confusion for some users as to which document is selected for use with their current query.
- This new selection mechanism could be used in the future to scale the number of document included in a search when/if the project moves to allow more than one document to be used in a single query.

- **Other information**:
**Components Added**
- PopupSelector - A popup which takes in a list of items and displays them in a select-like format.  The popup has a reletive position to its parent component and can be positioned on top of the screen as set by the popup's input parameters.
- ConversationSourceList - A pill list which renders a list of documents which have been selected by the user at the time of writing a query.  Currently, only one document is able to be selected at one time, however, this component is capable of handling more than one component if there is a need in the future.
- ConversationInputList - The input text box from the Conversation component refactored into its own component to handle all user input when writing a new query.

### DIFF
--- a/frontend/src/components/PopupSelector.tsx
+++ b/frontend/src/components/PopupSelector.tsx
@@ -1,0 +1,54 @@
+import React, { useRef } from 'react';
+import { useOutsideAlerter } from '../hooks';
+
+export function PopupSelector<SelectionItems extends { name: string }>({
+  isOpen = false,
+  setIsPopupOpen,
+  positionTop = 0,
+  positionLeft = 0,
+  headerText,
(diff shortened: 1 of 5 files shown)


siiddhantt:
Issue closed by #1733

PR #1039, opened 2024-07-19
Feature(Sharing): Introducing Promptable conversations 

ManishMadan2882:
### Backend
 1. Made POST `/api/share` endpoint compatible with promptable sharings
    * If the api key is existing with the same attributes(prompt_id, chunks, source), shared conversation depends on that api_key. New shared conversation doc is created, responding with the UUID as identifier.
    * else if the api key is existing and a shared conversation document with exactly same fields is already existing, avoid re-creating the shared_conversation and return the pre_existing UUID as identifier.
    *  generate a new api key, assign its key to the new shared conversation, respond with the new document's uuid.

2. Modified the GET `/api/shared_conversation/<string:identifier>` endpoint to send the api_key if present in response.
3. `complete_stream()` : Prevent storing request made through api_key

### Frontend
   - Structured the redux slice for sharedConversations.
Screenshot from 2024-07-26 15-08-57
   - Integrated the stream route, for prompting through shared conversations' api-keys. 
   - Further queries after the originally shared ones, will be saved in `localStorage` 
     Allow sharing conversations which can be prompted further


dartpain (approved):
All good, thank you!

PR #1048, opened 2024-08-01
Optimize Dockerfile to reduce image size and improve build efficiency

PeterDaveHello:

	Optimization of Dockerfile to reduce image size and improve build efficiency.

	This change aims to reduce Docker image size and enhance build efficiency. By merging apt-related RUN commands, we consolidate cleanup operations within the same layer, effectively reducing image size. Additionally, we removed the redundant apt-get clean command, as this operation is automatically handled in the Ubuntu base image.

- **Other information**:

	The reason why to remove `apt-get clean` is based on Docker's official best practices guide. More details can be found at:
	- https://docs.docker.com/build/building/best-practices/#apt-get

	These changes not only reduce image size but also improve the efficiency of the build process, positively impacting continuous integration and deployment workflows.

### DIFF
--- a/application/Dockerfile
+++ b/application/Dockerfile
@@ -4,14 +4,11 @@ FROM ubuntu:24.04 as builder
 ENV DEBIAN_FRONTEND=noninteractive
 
 RUN apt-get update && \
-    apt-get install -y software-properties-common
-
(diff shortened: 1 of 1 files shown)


dartpain (approved):
Thank you!

PR #1092, opened 2024-08-23
React widget: Sync with the Stream endpoint, error handling

ManishMadan2882:
    As the error messages are streamed in the /stream endpoint, it gets rendered in the query response to the user
     Suitable response to the user in case of error.


dartpain:
Thank you!

Issue #1106, opened 2024-08-28
🐛 Bug Report: CSV upload doesn't work


pabik:
### 📜 Description

CSV format upload doesn't work.

### 👟 Reproduction steps

1. Click on upload
2. Select .csv file

### 👍 Expected behavior

Name field should display the selected .csv file. After that user should be able to press train. 

### 👎 Actual Behavior with Screenshots

Name field stays empty. Train button is locked. 

### 💻 Operating system

MacOS

### 🤖 What development environment are you experiencing this bug on?

Docker

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

None

### 🧑‍⚖️ Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ruhi47:
Hi, I can help, have experience with GenAI Chatbots.
Please assign me, Thank you!

dartpain:
Appreciate your interest @ruhi47
Unfortunately this issue is fixed.
Please check out our other issues or suggest yours. 
Thank you!

PR #1109, opened 2024-08-31
CSV upload

aameraryan:

This PR addresses two issues:

1. Fixes the [CSV file upload issue](https://github.com/arc53/DocsGPT/issues/1106).
Additional info and screenshots are attached in the bottom.

2. Resolves a linting issue caused by the incompatibility between prettier 3.3.3 and eslint-plugin-prettier 4.2.1. This is a known [issue](https://github.com/prettier/eslint-plugin-prettier/issues/562). Upgrading eslint-plugin-prettier to version 5.x resolved the problem. The upgrade also required formatting the files to pass lint checks.

Note: The GitHub action was blocking commits, so the fix for eslint-plugin-prettier and the related updates were included in this PR. Please let me know if we should open a separate issue for this fix and address it before tackling the main issue.

- **Why was this change needed?** ([You can also link to an open issue here](https://github.com/arc53/DocsGPT/issues/1106))

- **Other information**:
Before: The file input did not support uploading .csv files. Even when users selected a .csv file, the file input did not select the file and appeared empty.
Screenshot 2024-08-31 at 5 24 24 AM

After: The file input now supports and successfully uploads .csv files.
Screenshot 2024-08-31 at 4 04 01 AM

Screenshot 2024-08-31 at 4 05 24 AM

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -37,7 +37,7 @@
         "eslint-config-standard-with-typescript": "^34.0.0",
         "eslint-plugin-import": "^2.27.5",
         "eslint-plugin-n": "^15.7.0",
-        "eslint-plugin-prettier": "^4.2.1",
+        "eslint-plugin-prettier": "^5.2.1",
         "eslint-plugin-promise": "^6.6.0",
         "eslint-plugin-react": "^7.35.0",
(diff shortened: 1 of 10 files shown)


dartpain:
Perfect, thank you so much!

dartpain (approved):

PR #1113, opened 2024-09-04
Refining the scrolling functionality while streaming

ManishMadan2882:
     * As per the current logic, the `endMessageRef` was referred while scrolling
     * In this PR, referring to the `conversationRef` alone to control scrolling
     * The `scrollIntoView` method should handle the following cases:
         - When the status is `idle` or a new  prompt is just added: Smoothly scroll to the bottom of `conversationRef`
         - While streaming: when the response is getting generated, the scroll should stick to the bottom of the conversation container, to make the streaming more consistent.
    * Frictionless streaming.
- **Other information**:
  Fixed the overflow of code, without any language match.
  Minor change in react-widget/publish.sh : Refresh the package json and package-lock after each publish

### DIFF
--- a/extensions/react-widget/package-lock.json
+++ b/extensions/react-widget/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "docsgpt",
-  "version": "0.4.1",
+  "version": "0.4.2",
   "lockfileVersion": 3,
   "requires": true,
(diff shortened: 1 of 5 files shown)


dartpain (approved):
All good, thanks!

PR #1115, opened 2024-09-07
support for lancedb as vectordb

akashAD98:
Added support for  lancedb as vector store

### DIFF
--- a/application/core/settings.py
+++ b/application/core/settings.py
@@ -18,7 +18,7 @@ class Settings(BaseSettings):
     DEFAULT_MAX_HISTORY: int = 150
     MODEL_TOKEN_LIMITS: dict = {"gpt-3.5-turbo": 4096, "claude-2": 1e5}
     UPLOAD_FOLDER: str = "inputs"
-    VECTOR_STORE: str = "faiss"  # "faiss" or "elasticsearch" or "qdrant" or "milvus"
+    VECTOR_STORE: str = "faiss"  # "faiss" or "elasticsearch" or "qdrant" or "milvus" or "lancedb"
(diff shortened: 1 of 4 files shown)


akashAD98:
@dartpain can you please check it

application/core/settings.py
dartpain, line 1: Im not sure if in memory is the best approach here as we have different containers that need to access lancedb. Does it work when you upload docs?

application/vectorstore/lancedb.py
```diff
@@ -0,0 +1,61 @@
...
+                pa.field("vector", pa.list_(pa.float32(), list_size=len(vectors[0]["vector"]))),
+                pa.field("text", pa.string()),
+                pa.field("metadata", pa.map_(pa.string(), pa.string())),
+            ])
+            self.table = self.db.create_table(self.table_name, schema=schema)
+
+        # Add data to the table
+        self.table.add(vectors)
+
+    def search(self, query: str, k: int = 2, *args, **kwargs):
```
dartpain, line 52: Please make sure that you filter documents using self._path

application/vectorstore/lancedb.py
```diff
@@ -0,0 +1,61 @@
+
+
+
+from application.vectorstore.base import BaseVectorStore
+from application.core.settings import settings
+import lancedb
```
dartpain, line 6: Lets lazy load it inside __init__

dartpain (changes requested):
Thank you so much for creating this PR. really appreciate that its using native library too!
Requested a few minor change and happy to merge after.

application/vectorstore/lancedb.py
akashAD98: yes please check it

akashAD98:
i created new pr so closing this

Issue #1137, opened 2024-09-19
🐛 Bug Report: Running Docker on local machine doesn't reflect FE changes


anasKhafaga:
### 📜 Description

When I installed the echo-system, I have used `docker-compose-local.yaml` file as I prefer to quickly start the environment via docker. Unfortunately I discovered that there's no bind mount to the `frontend/src/` directory outside the container world. This issue is solved in #1138 

### 👟 Reproduction steps

1. Hit
```bash
$ docker compose -f docker-compose-local.yaml up
```

2. Try to make a change in `frontend/src` directory.
3. Check your browser to see the change

### 👍 Expected behavior

Change should reflect immediately without the need to rebuild the image.

### 👎 Actual Behavior with Screenshots

There's no reflection happens unless you re-build the image and re-run the container.

### 💻 Operating system

Linux

### 🤖 What development environment are you experiencing this bug on?

Docker

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

Yes, I am willing to submit a PR!

### 🧑‍⚖️ Code of Conduct

- [X] I agree to follow this project's Code of Conduct

PR #1138, opened 2024-09-19
fix(Docker): Integrate a bind mount in the local docker compose file

anasKhafaga:
- Bug fix related to issue #1137 

This change enhances FE development experience using Docker.


dartpain:
Thank you!

PR #1153, opened 2024-09-26
fix: updated docker execution script from `docker-compose` to `docker…

utin-francis-peter:
… compose` so it works on latest versions of docker used across contributors machine

Bug fix

It's very likely the majority has migrated to Docker V2. The update is required for running docker files in latest version.

- **Other information**:


dartpain (approved):
Thank you!

Issue #1169, opened 2024-09-30
Add a default message in the UI when no data is fetched, Example: Settings > Chatbots - No existing Chatbots.


dartpain:
Need to handle such empty state and prompt users for action in:

- Settings -> documents menu
- Settings -> chatbots menu

anasKhafaga:
Could you please assign this issue to me @dartpain ?

githubashutoshsoni:
Hi, can you please assign this issue to me?

PR #1178, opened 2024-10-01
Feat(1169): Handle state of empty tables

anasKhafaga:
  Feature update: Added a "No existing ..." messages in multiple locales (English, Spanish, Japanese, and Chinese), and updated both the `APIKeys.tsx` and `Documents.tsx` components to display a message when there are no API keys or documents available.

  related to #1169. This change was needed to improve user experience by providing feedback when there are no existing chatbots or documents in their respective sections, ensuring users are clearly informed when no data is available across multiple languages.

- **Other information**:  
image

### DIFF
--- a/frontend/src/locale/en.json
+++ b/frontend/src/locale/en.json
@@ -62,7 +62,8 @@
       "name": "Name",
       "key": "API Key",
       "sourceDoc": "Source Document",
-      "createNew": "Create New"
+      "createNew": "Create New",
+      "noData": "No existing Chatbots"
     },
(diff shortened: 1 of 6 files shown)


ManishMadan2882 (changes requested):
Hi @anasKhafaga
Requesting a minor change, there is slight border overlap probably because of `border-r` class in the td
image
In the right bottom the two borders are overlapping.
Thankyou!

ManishMadan2882 (approved):
Thanks @anasKhafaga !

PR #1216, opened 2024-10-04
 sharepoint_loader.py

sOnU1002:
pip install Office365-REST-Python-Client
 run this one


dartpain:
Make sure that you also make relevant changes in the frontend such that users can sync their file/ folder.
Also changes in the remote_creator file and worker.py file too

Thank you!

PR #1217, opened 2024-10-04
Update docs_parser.py

sOnU1002:
pip install PyPDF2 pytesseract pdf2image pillow


dartpain:
This seems to be a duplicate

PR #1238, opened 2024-10-04
Bug Fix: Crash, abnormal update while switching conversation Ids

ManishMadan2882:

    Basically conversationId is updated in two situations:
       1. switching the conversation
       2. streaming when data.type = 'id'

     As per the current flow, there are following bugs:

- While streaming, switching to New Chat, crashes the application
- In case of New Chat, the conversations are not updating in the sidebar

     Fix:
- Pause the update of streaming query without pausing the fetchAnswer middleware explicitly.
             * Set the status as idle for every update to conversationId
             * Append the streaming answer to the response only when the status is not idle.
        - Fixes the root cause of the problem.


dartpain (approved):
Thank you!
LGTM!

PR #1240, opened 2024-10-05
Update web_loader.py

sOnU1002:


dartpain:
Hey @sOnU1002 
I appreciate you commits, I really do. And it looks like you are doing some things by accident.
By I have to warn you to be careful with your PR's.
We have too many now, most of them are incomplete.
Please also consider creating a draft pull requests. And fill in the for on PR submission

PR #1299, opened 2024-10-11
Feat: Add ability to start docker if it's not running

### LINKED ISSUE #1298: 🚀 Feature: Add function to check if docker daemon is running in setup.sh script
### 🔖 Feature description

`docker info` returns an exit code of 0 if the Docker daemon is not running.

We can utilise this to check if docker is running before calling `docker-compose` commands. Although platforms like Linux spin up a daemon if it's not running, it's not the case for macOS or Windows.

Better developer experience

- Update setup.sh script to add a new function `check_and_start_docker`
- The new function will check if docker is not running and attempt to start it and wait while docker info doesn't return success status

### 🔄️ Additional Information

_No response_

shatanikmahanty:
Feature to start docker daemon if not running

Fixes #1298

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -9,6 +9,54 @@ prompt_user() {
     read -p "Enter your choice (1, 2 or 3): " choice
 }
 
+check_and_start_docker() {
+    # Check if Docker is running
+    if ! docker info > /dev/null 2>&1; then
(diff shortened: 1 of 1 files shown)


dartpain (approved):
LGTM, thank you!

PR #1321, opened 2024-10-14
refactor(user/routes): simplify parsing of isPromptable

Juneezee:

Refactoring

We can use the `inputs.boolean` from flask-restx [^1] to parse boolean for us. This way `is_promptable` is a `bool` variable and not a `str` variable so we don't need to do `.lower() == "true"` everywhere.

[^1]: https://flask-restx.readthedocs.io/en/latest/api.html?highlight=boolean#flask_restx.inputs.boolean

- **Other information**:


ManishMadan2882 (approved):
Thanks @Juneezee !

dartpain:
@holopin-bot @Juneezee  Thank you!

Issue #1229, opened 2024-10-04
🐛 Bug Report: Can’t scroll chat, browser tries to refresh page


pabik:
### 📜 Description

Sometimes on iOS Safari (probably all iphone) chat gets stuck, as you can't scroll up browser tries to refresh. 

### 👟 Reproduction steps

This doesn't happen every time, but on iOS start a new chat and try to scroll up.

### 👍 Expected behavior

You should be able to scroll chat freely in any direction. 

### 👎 Actual Behavior with Screenshots

![image](https://github.com/user-attachments/assets/5046e98e-9bf4-458b-89bc-9fdae919450d)
![image](https://github.com/user-attachments/assets/95fd00c5-3ca3-4507-a29c-c11e505c8079)

### 💻 Operating system

Something else

### What browsers are you seeing the problem on?

Safari

### 🤖 What development environment are you experiencing this bug on?

Docker

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

None

### 🧑‍⚖️ Code of Conduct

- [X] I agree to follow this project's Code of Conduct

albinsabu2023:
hey can you assign this to me i can work on this

Vaishnavi-Raykar:
I would like to work on it !

SaumyaSubham:
I want to work on this issue and Kindly provide some more info to get me started!
And Assign me under hacktoberfest 2024

SaumyaSubham:
Hii @dartpain ,
Could you please locate the file that I have to work so that I can get access easily and start working on it! Waiting for your reply.
Thank You

ManishMadan2882:
@SaumyaSubham the file you are looking for should  be `./frontend/src/conversation/Conversation.tsx`
This issue is related to iOS and Safari specifically, so make sure you that you are able reproduce this issue.

prathamesh424:
@ManishMadan2882  , @dartpain  can you assign this issue to me

ManishMadan2882:
@nayelimdejesus Do you think this issue still exists? 
Also, if there is any specific device you are able to reproduce this issue on please share

nayelimdejesus:
I was trying to look into it just now and it was actually working just fine for me.

ManishMadan2882:
Thanks for the confirmation @nayelimdejesus 
Closing the issue for now

PR #1322, opened 2024-10-14
Fix iOS Safari Scrolling Issue in Conversation Component #1229

SaumyaSubham:
### Description
This PR addresses an issue #1229 specific to iOS Safari where users were unable to scroll through the chat smoothly. The issue caused the chat to get "stuck" while attempting to scroll up, and the browser would sometimes try to refresh the page due to the pull-to-refresh feature on iOS.

### Reproduction Steps:

- On iOS Safari (tested on iPhone devices), start a new chat.
- Attempt to scroll up in the chat window. In some cases, scrolling up triggers a refresh of the page.

**Expected Behavior:**
Users should be able to scroll up and down within the chat freely without the page attempting to refresh.

**Actual Behavior:**
Scrolling up sometimes gets stuck, and the browser tries to refresh the page.

### Solution:
- Event Listener for touchmove: Added a useEffect hook to listen for the touchmove event and prevent the default pull-to-refresh behavior when the user reaches the top or bottom of the scrollable chat area. This is done using e.preventDefault() on iOS devices to disable the default action.

- Check Scroll Position: The logic ensures that the prevention only applies when the scroll position is at the very top or very bottom of the chat content. This ensures users can still scroll normally within the chat, but the refresh action is blocked in these edge cases.

### Testing:
I directly tested this fix on BrowserStack(browser simulator). I would kindly request the maintainers or other collaborators to verify this fix on an actual iOS device (Safari browser).

Thank You and I'm waiting for your feedback.

### DIFF
--- a/frontend/src/conversation/Conversation.tsx
+++ b/frontend/src/conversation/Conversation.tsx
@@ -44,6 +44,7 @@ export default function Conversation() {
   const handleUserInterruption = () => {
     if (!eventInterrupt && status === 'loading') setEventInterrupt(true);
   };
+
   useEffect(() => {
     !eventInterrupt && scrollIntoView();
   }, [queries.length, queries[queries.length - 1]]);
@@ -58,7 +59,7 @@ export default function Conversation() {
   useEffect(() => {
     if (queries.length) {
       queries[queries.length - 1].error && setLastQueryReturnedErr(true);
-      queries[queries.length - 1].response && setLastQueryReturnedErr(false); //considering a query that initially returned error can later include a response property on retry
+      queries[queries.length - 1].response && setLastQueryReturnedErr(false); // considering a query that initially returned error can later include a response property on retry
     }
   }, [queries[queries.length - 1]]);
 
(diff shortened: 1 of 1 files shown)


SaumyaSubham:
Kindly look up to the pull request that I have provided.

dartpain:
hey @SaumyaSubham 
Seems like there is a build error
src/conversation/Conversation.tsx(282,12): error TS2741: Property 'handleQuestion' is missing in type '{}' but required in type '{ handleQuestion: ({ question, isRetry, }: { question: string; isRetry?: boolean | undefined; }) => void; }'.
src/conversation/Conversation.tsx(290,19): error TS2322: Type '"PROMPT"' is not assignable to type 'MESSAGE_TYPE'.
src/conversation/Conversation.tsx(299,15): error TS2322: Type 'Element' is not assignable to type 'string'.
Error: Command "npm run build" exited with 2

Thank you!

SaumyaSubham:
@dartpain Thank You for the suggestion. I'm looking up to it!

PR #1330, opened 2024-10-15
Fixes - 1292

### LINKED ISSUE #1292: 🚀 Feature: New button and logo in header when the sidebar is closed
### 🔖 Feature description

The "+" icon will open a new chat on click. (Only if the current chat has any messages)

https://www.figma.com/design/OXLtrl1EAy885to6S69554/DocsGPT?node-id=0-1&t=UbTE0eKgODqCaJ8p-1

image

image

image

image

Better UI/UX

Better UI/UX

### 🔄️ Additional Information

_No response_

None

Niharika0104:
Fixes #1292

Made UI/UX changes.

image
image
image


ManishMadan2882 (changes requested):
Hi @Niharika0104 
Thanks for the update, just reviewed.
The New Chat Icon is visible on the chat with no conversations in desktop screen.
Also, could you please sync your branch with the upstream main.
image

Niharika0104:
@ManishMadan2882 
I incorporated the requested changes and also updated the branch with upstream branch
image
image

ManishMadan2882 (changes requested):
Hi @Niharika0104 
I closely monitored the Redux States, and found that there is one case which will need improvement:
Try clicking the new chat button while one Conversation is streaming.
Noticed that the queries are emptied at this point,status is set to idle but the `conversationId` is not set to null.
Could you please look into this case and submit a fix.
Thanks!

Niharika0104:
@ManishMadan2882 I'm not clear with your exact ask
 when some conversation is loading(loading icon in the input box) and during that time if user clicks on new Chat button then conversation id should be set to null but it is not happening
Is it what you are asking me or  did i get wrong?

ManishMadan2882:
Exactly @Niharika0104

Niharika0104:
@ManishMadan2882 
image
I see the similar issue even when i click on the new chat button in the navbar panel 
So the issue is already there from the beginning right ?

Niharika0104:
I just mistakenly clicked on close with comment

ManishMadan2882:
I think the conversationId should be null as we move to a fresh conversation, because there are few things dependent on this functionality, Like the selection in the sidebar is not updated, the Share icon shouldn't be there for a null ConversationID.
So, yeah it is necessary that the conversationId should be null in any case we move to a new chat.

image
You can see in this screenshot that the Share icon on top right, the sidebar selection are inconsistent, think it must be a minor fix.

Niharika0104:
@ManishMadan2882 are you saying that when we click on a particular conversation slice we are unable to load the queries present in those conversation id on the conversation screen?
image

ManishMadan2882:
No, you can reproduce the issue by these steps:
* Prompt the conversation
* By the time the conversation loads, click the New Chat (it is still loading/streaming)
* When you come to new chat, you will notice that the:
      - ConversationId hasn't updated although the queries are emptied
      - The Share button is still visible as it is rendered where the conversationId is not null
      - The sidebar is not getting updated, because the conversationId is not updated
The conversationId is not updated to null only the queries have been emptied which is the problem.

Niharika0104:
@ManishMadan2882 Can you check it now?
The issue is resolved now.

ManishMadan2882 (approved):
Looks Good, Thanks @Niharika0104 !

dartpain:
@holopin-bot @Niharika0104  Thank you!

PR #1345, opened 2024-10-19
Bug fix: /api/share route 

ManishMadan2882:

  * Steps to reproduce this bug:
         1. Create a conversation
         2. Create a shared link for the conversation, with enable allow users to prompt further
         3. Create another shared link for the conversation, this time disable the option to prompt further
     
  * Current behaviour: The same uuid is assigned to both the shared conversations 
  * Expected behaviour: A shared conversation which is promptable with a particular doc should be unique avoiding redundant entries; while a non-promptable shared conversation should also be unique but distinct from a promptable one

     Avoids ambiguity in the behaviour of shared conversations.

### DIFF
--- a/application/api/user/routes.py
+++ b/application/api/user/routes.py
@@ -939,7 +939,7 @@ class ShareConversation(Resource):
                     "conversation_id": DBRef(
                         "conversations", ObjectId(conversation_id)
                     ),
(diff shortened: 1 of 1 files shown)


dartpain (approved):
LGTM, thank you!

Issue #1310, opened 2024-10-13
🐛 Bug Report: Basic UI changes


SamsShow:
### 📜 Description

Noticed some basic UI issues which can affect the user experience 

1) The page title contains an emoji [🦖](https://emojipedia.org/t-rex) which seems a little odd 
2) The user profile picture will look much better than the 💻🧑 emoji
3) Source resources should come after the answer not before the answer 
4) Lack of proper spacing/padding in mobile devices 

### 👟 Reproduction steps

Go to https://app.docsgpt.cloud/

### 👍 Expected behavior

2) ![image](https://github.com/user-attachments/assets/6e418483-5e2c-473f-bbca-a8f269d7c390)
3) ![image](https://github.com/user-attachments/assets/20b6cc97-5483-499d-afd1-dd7ab32e9094)

### 👎 Actual Behavior with Screenshots

1) ![image](https://github.com/user-attachments/assets/0c77c364-5225-4d7b-a19e-f53282106c0b)
2) ![image](https://github.com/user-attachments/assets/ad62fcf0-f90c-4c84-bd99-ec1a80e7833e)
3) ![image](https://github.com/user-attachments/assets/aa002037-aef3-4c02-ab1d-906d2e51f831)
4) ![image](https://github.com/user-attachments/assets/dbf2ad0c-c495-4dab-91e2-e681b0d0913e)

### 💻 Operating system

Windows

### 🤖 What development environment are you experiencing this bug on?

Docker

### 👀 Have you spent some time to check if this bug has been raised before?

- [X] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

Yes, I am willing to submit a PR!

### 🧑‍⚖️ Code of Conduct

- [X] I agree to follow this project's Code of Conduct

SamsShow:
I would like to work on this issue

dartpain:
In not sure that we want to change 3. as it was an intentional change to provide sources while answer is still streaming. In terms of 
1 - thats ok
2- I agree
4. Please check current issues to fix mobile things, but yeah I dont remember anyone trying to improve padding.

Thank you! Nice to see you here again!

SamsShow:
[...]

PR #1383, opened 2024-10-26
fix: Enhance layout in Conversation components

SamsShow:
fixes: 
1) 
image
2)
image
4)
image

#1310


ManishMadan2882 (changes requested):
Thanks @SamsShow 
Could you please check the avatar for mobile responsiveness.
image
Think it would be better if placed like the dino icon next to Answer.

SamsShow:
@ManishMadan2882 
image
was working fine for a short message that's why I missed it. Will make the necessary changes

SamsShow:
@ManishMadan2882 made the changes

ManishMadan2882 (changes requested):
@SamsShow 
Sorry for not being clear before, only wanted the icon to be positioned like that on mobile screens, also the `Prompt` text is not required.
Just need a minor change to make the Avatar align as it was before 
* Small message
image
* Large message
image

PR #1384, opened 2024-10-26
Fix: empty doc selection in /stream

### LINKED ISSUE #948: 🐛 Bug Report: Empty document uses default document
### 📜 Description

When Empty document is selected documents from pre-uploaded DocsGPT documentation (Default option) is used. 

### 👟 Reproduction steps

1. Choose Empty from Source Doc selector.
2. Ask any question, you will see that DocsGPT still uses Default source doc. 

### 👍 Expected behavior

Upon selecting Empty, model should not receive any documents. 
The behaviour should be the same as if Chunks processed per query is set to 0. 

Model receives Default documents. 

### 💻 Operating system

Linux

_No response_

Docker

_No response_

_No response_

### 📖 Relevant log output

_No response_

None

### 🧑‍⚖️ Code of Conduct

- [X] I agree to follow this project's Code of Conduct

ManishMadan2882:
     * Purged selectedDocs from /stream endpoint
     * Chunks will be set to zero in case of isNoneDoc (Empty).
        Fixes #948


dartpain (approved):
Lgtm

Issue #1341, opened 2024-10-18
Feature: Backstage widget plugin


dartpain:
Create a backstage plugin that will utilise our existing react widget.
Check out examples here:
https://backstage.io/plugins/

Import our react extension that is here:
[react web extension](https://github.com/arc53/DocsGPT/tree/main/extensions/react-widget)

https://www.npmjs.com/package/docsgpt

This issue will need its own PR, but is a subtask of #1293

ayaan-qadri:
I would like work on it !

ayaan-qadri:
I would like to work on it !

ayaan-qadri:
(WORKING)

naaa760:
Hello @dartpain  :)
I want to contribute to this issue. Please assign this issue to me.
Thank you!

ayaan-qadri:
Update ::: making changes in react-widget for this issue.

ayaan-qadri:
Update ::: Custom sizing for <DocsGPTWidget /> in progress

ayaan-qadri:
This issue is resolved.

PR #1389, opened 2024-10-28
Added custom size and Default widget open in <DocsGPTWidget />

ayaan-qadri:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...) <br/>
      1. Allow to custom size of DocsGPTWidgte using height, width, max width, max height
      2. Added defaultOpen prop which was necessary to open widget when docsgpt-widget-plugin route hits without explicitly clicking on widget icon <br/><br/>
- **Why was this change needed?** (You can also link to an open issue here)<br/>
     These changes offer support for the sub problem of issue #1341. <br/><br/>
- **Other information**: <br/>
     Custom sizing does not work in < 1280px screen (as far as i know, i do not think that backstage going to be use in mobile / < 1280px), lemme know if we need to add custom size prop also affect screen sizes < 1280px


ManishMadan2882 (commented):
@ayaan-qadri 
Other components using WidgetProps won't align with the custom object for size
image
Size is ideally string in the min-width and height

ayaan-qadri:
@ManishMadan2882 
image

So I'm making changed into height and width of prompt window only,
I also changed into interface for size, In the current implementation it will not affect 'StyledButton' or 'PromptContainer' as they are taking size parameter even if someone passes custom object as argument (I also tested myself, it is working as before)

If we make those styled-component size prop's type string (as it was before) so it throws Linting error

In my opinion, making size prop compatible custom object is more reliable and future proof as well: 

(Lemme know your thoughts that what should we do for this as if the current implementation does not fits code style)

extensions/react-widget/src/components/DocsGPTWidget.tsx
```diff
@@ -265,7 +295,7 @@ const DotAnimation = styled.div`
 const Delay = styled(DotAnimation) <{ delay: number }>`
   animation-delay: ${props => props.delay + 'ms'};
 `;
-const PromptContainer = styled.form<{ size: string }>`
```
ManishMadan2882, line 268: @ayaan-qadri I was thinking if you could just replace the need of passing the sizes to different styled components.
Having an object of dimensions for small, medium, large and the custom.
So that the width, height could be controlled from the parent directly.

ayaan-qadri:
@ManishMadan2882 Please give me some time, I will sure look into this.

extensions/react-widget/src/components/DocsGPTWidget.tsx
ayaan-qadri: Now we can control dimensions from sizesConfig.

Issue #1231, opened 2024-10-04
🚀 Feature: Modals under one style


pabik:
### 🔖 Feature description

Currently in an app all modals look different. I think it will benefit the UI to have them under one style.
Create new modal in settings/chatbots looks clean and I think should be the base case.
![image](https://github.com/user-attachments/assets/945c85e2-9c2d-4c31-af12-824e1f320dae)
However I think an off modal click/tap should close modal. 

Here for example what would need to be changed in upload modal.
![image](https://github.com/user-attachments/assets/08fff4c7-ef73-47af-880c-55a53cebf6de)
1. Add x at the top right corner to close. 
2. Remove 'Cancel'
3. Make 'Train' button inactive by default, similarly to the one in new chatbot modal.
![image](https://github.com/user-attachments/assets/984442fd-e975-46a3-8ed5-cb6b1c29d09f)
![image](https://github.com/user-attachments/assets/4bcc4916-8773-4fc0-9385-5694f5dde9ef)
4. Allow for a click outside modal to close it. 

### 🎤 Why is this feature needed ?

Benefit UI. 

### ✌️ How do you aim to achieve this?

1. Add click outside modal to close it functionality.
2. Check other modals to have same structure.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

None

JeevaRamanathan:
I noticed that another issue,  [Updated UI/UX for "Upload new document" modal](https://github.com/arc53/DocsGPT/issues/1210)  is already in place. Could having multiple issues focused on updating the upload modal lead to some overlap?

dartpain:
Think this one is more about the color of the buttons for consistency.
Other issue is a larger change and we dont expect it to be delivered quickly as this one can.
I do appreciate your careful sift through other issues

Kritika30032002:
assign this issue to me, interested to contribute @pabik @dartpain

uttam282005:
@dartpain  can you assign me this issue. I have explored the codebase and files related to this issue and I think I can work on this.

Kritika30032002:
working on it

prathamesh424:
@ManishMadan2882  can you assign this issue to  me

AkashJana18:
Hey  @dartpain @ManishMadan2882 If this issue still persist I would like to work on it. Please assign it to me. Thank you.

RohittCodes:
[...]

PR #1434, opened 2024-11-12
feat: wrapper modal

RohittCodes:
this pr introduces a wrapper modal, which will act as a base for every other modals. includes functionality to close the modal on a click outside the modal or an `ESC` key press.

fixes: #1231 

- **Other information**:

### DIFF
--- a/frontend/src/Navigation.tsx
+++ b/frontend/src/Navigation.tsx
@@ -32,7 +32,6 @@ import {
   selectConversations,
   selectModalStateDeleteConv,
   selectSelectedDocs,
-  selectSelectedDocsStatus,
   selectSourceDocs,
   selectPaginatedDocuments,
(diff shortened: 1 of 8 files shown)


ManishMadan2882 (changes requested):
Thanks @RohittCodes 
Looks good, requesting one minor change to let the train button disabled for remote uploads as well, when the link field is empty
image

RohittCodes:
@ManishMadan2882 should I try adding something to prevent the modal from closing while training any upload?

ManishMadan2882:
@RohittCodes Yes, we should avoid closing of the modal while upload/train is in process, good catch!

ManishMadan2882 (approved):
LGTM, thanks @RohittCodes !

Issue #1166, opened 2024-09-30
[BUG] Make sure users can only send feedback once on convo and make it persistent


dartpain:
Basically both on a normal conversation page and on a shared conversation page its sometimes possible to submit feedback twice after refreshing / reloading the page. It good to make feedback persistend

sidhanshamil:
I would like to work on this

anurag-4508:
Is this issue still open? I want to work on it. Please assign me this task.
Thank you

Niharika0104:
@dartpain Can you assign this issue to me?

PR #1455, opened 2024-11-26
Fixed the feedback issue

Niharika0104:
Fixes - #1166


ManishMadan2882 (changes requested):
Hi @Niharika0104 
There is an error, could you please resolve.
```
line 201
    {"_id": ObjectId(data["conversation_id"]), f"queries.{data["question_index"]}": {"$exists": True}},
                                                                ^^^^^^^^^^^^^^
SyntaxError: f-string: unmatched '['
```

Niharika0104:
@ManishMadan2882 can you check it now?

ManishMadan2882 (changes requested):
@Niharika0104 
Thanks for the update, I have few points:
* The feedback once submitted, is not getting deleted. Like I liked once, but need to remove my reaction.
* The feedback once submitted, removes the other option. Like if I liked once, the dislike button is not visible.
image

Niharika0104:
> @Niharika0104 
> Thanks for the update, I have few points:
> * The feedback once submitted, is not getting deleted. Like I liked once, but need to remove my reaction.
> * The feedback once submitted, removes the other option. Like if I liked once, the dislike button is not visible.
> image
> 

I thought that once the feedback is given it shouldn't be taken again.. that's how it's happening in the chatGPT as well

Niharika0104:
@ManishMadan2882 so how should we change this and we want to take feedback only once the issue desc says so...

Niharika0104:
> If the user clicks on the reaction which they have already submitted, it should get removed from the DB.

so when I click submitted feedback, then that particular feedback should be deleted from the db and also we should show the like and unlike buttons too.. 
correct me if I'm wrong..

ManishMadan2882:
> > If the user clicks on the reaction which they have already submitted, it should get removed from the DB.
> 
> so when I click submitted feedback, then that particular feedback should be deleted from the db and also we should show the like and unlike buttons too.. correct me if I'm wrong..

Yep @Niharika0104

ManishMadan2882:
@Niharika0104 are you active on this PR?

ManishMadan2882:
@Niharika0104 I think you have not pushed the changes.

Niharika0104:
@ManishMadan2882 Can you please check it now?

ManishMadan2882 (approved):
@Niharika0104 LGTM, Thanks!

PR #1456, opened 2024-11-26
Minor UI adjustments

ManishMadan2882:
Minor UI adjustments:
     -  Code used to overflow in the conversation bubble when there was no match for language
     -  The input field in Conversation touched the bottom of the viewport in mobile screens.
     -  Font size in the user questions looked larger in mobile browsers.
     - Limited characters in Settings/Docs in table column document name to avoid overflows
     - In `api/delete_old` remove docs from the collection even if they are missing from the file tree.
     
image

      UI


dartpain (approved):
Great, thank you!

PR #1459, opened 2024-11-27
feature: drag and train functionality.

### LINKED ISSUE #1167: Drag and drop training, from input box
Basically should trigger from file training process once a user drags and drops a file into the input box

Srayash:
feature: drag and train functionality.
fixes #1167

### DIFF
--- a/frontend/src/Navigation.tsx
+++ b/frontend/src/Navigation.tsx
@@ -476,6 +476,7 @@ export default function Navigation({ navOpen, setNavOpen }: NavigationProps) {
       />
       {uploadModalState === 'ACTIVE' && (
         <Upload
+          recievedFile={[]}
           setModalState={setUploadModalState}
           isOnboarding={false}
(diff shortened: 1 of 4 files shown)


Srayash:
Currently, this logs the dropped file to the console, gotta add the upload functionality but I'm stuck as the conversation and upload are 2 different components. maybe a new redux slice has to be implemented.

ManishMadan2882:
Hi @Srayash 
Think we can do it in this way, by importing the Upload component in the conversation.tsx
Certain changes will be required in the Upload component, where the activeTab is by default null we can maybe make a prop of the activeTab's default value lets say "file" for our use case.
But still the issues might arise when passing files as prop from the conversation component, because the states are already set as the component gets mounted, so better conditionally render the component when modalState === 'ACTIVE'

Srayash:
Some tests are failing after I merged main into this branch, will close this for now and raise a new PR.
thank you.

PR #1461, opened 2024-11-27
(fix:edit-query)delete following queries

ManishMadan2882:
    1. Deletes following queries from the database
    2. updates the state safely in the frontend.
    3. Exacted to the UI design
    4. Close the editable input when clicked outside or <kbd>ESC</kbd>
     UX


dartpain (approved):
All good, thank you!

PR #1462, opened 2024-11-28
GenAI eatting image from DOCX

fadingNA:

> Chage from docx2txt.process to extract them manually
> Eatting image as Base64 | Table as HTML tag | Text for paragraph

> To using multiple vector store to retrieve correct image as paragraph order instead of zip them by docx2text.process()

- **Other information**:
- Progress the issue is now when we retrieve need to find the way to convert base64 back for AI to understand image
Screenshot 2024-11-28 at 12 00 19 PM

@dartpain  let me know if this correct approach or wrong direction

### DIFF
--- a/application/parser/file/docs_parser.py
+++ b/application/parser/file/docs_parser.py
@@ -3,12 +3,20 @@
 Contains parsers for docx, pdf files.
 
 """
-from pathlib import Path
-from typing import Dict
 
+from pathlib import Path
+from typing import Dict, List, Any, Union
+from base64 import b64encode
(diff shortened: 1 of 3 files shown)


fadingNA:
@dartpain this is the progress as we discussed on discord, let me know if this right direction.

- for the text embedding seems work properly except the default document that using 768 dimension but clip using 512 dimension. 
- Image maybe I have to resize to make base64 smaller before embed to CLIP model
- Issue here it stuck when I upload docx to server it seems stuck with document that contains images.

PR #1309, opened 2024-10-12
(i18n): Added support for Russian Language

kaleaditya779:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...)
- (i18n): Added support for Russian language. So that now russian speaking users can use this functionality.

- **Why was this change needed?** (You can also link to an open issue here)
- To help the software in localizing more. This gives more benefit to software when it comes to reaching to Russian speaking user base.

- **Other information**:
- None

PR #1486, opened 2024-12-11
Fix language support for Chinese and add Russian language

Srayash:
Traditional Chinese `zh-TW` translation file was present, but the language was missing from the language selection options.
Adds support for Russian language ( from #1309 )

- **Other information**:


dartpain (approved):
Thank you!

PR #1570, opened 2025-01-13
Widget fixes

ManishMadan2882:
       1. Center the Toolkit in the Search button
       2. Fixes the publish.sh -  package json is not affected my publish. Only version should be updated
      generally while publishing the targets were removed from the package json which may get committed and required a fix


dartpain (approved):
Thank you!

Issue #1585, opened 2025-01-16
🚀 Feature: enabling sound streaming through websockets with elevenlabs


aalghooneh:
### 🔖 Feature description

Enabling websocket streaming with the eleven labs api.

### 🎤 Why is this feature needed ?

to improve speed and efficiency, especially for long answers!

### ✌️ How do you aim to achieve this?

by leveraging the documentation on eleven labs websocket api

### Discussed in https://github.com/arc53/DocsGPT/discussions/1542

<div type='discussions-op-text'>

<sup>Originally posted by **aalghooneh** January  5, 2025</sup>
Hi,

I took a look at the project and really loved what you guys are doing here, really like to join you guys.

For this idea, I think the TTS can be a lot faster with the websockets for the eleven labs</div>

In reference to this discussion, I have made the necessary changes to my fork for this, on which branch should I do the PR ?

thanks 

### 👀 Have you spent some time to check if this feature request has been raised before?

- [X] I checked and didn't find similar issue

### Are you willing to submit PR?

Yes I am willing to submit a PR!

PR #1586, opened 2025-01-16
websocket implementation of elevenlabs

aalghooneh:
- **What kind of change does this PR introduce? feature

- **Why was this change needed? to improve speed for long responses

- **Other information**:[#1585](https://github.com/arc53/DocsGPT/issues/1585)


aalghooneh:
@dartpain could you give this a look?

ManishMadan2882 (commented):
@aalghooneh there is an error in importing base class in tts/elevenlabs.py, just check how we import it in GoogleTTS.

aalghooneh:
> @aalghooneh there is an error in importing base class in tts/elevenlabs.py, just check how we import it in GoogleTTS.

It should be okay now, could you check again ?

ManishMadan2882 (changes requested):
@aalghooneh After updating the endpoint /api/tts to use the ElevenlabsTTS, I get this error on the frontend
`{
  "error": "cannot unpack non-iterable NoneType object",
  "success": false
}
`

The output.mp3 is also not playable

aalghooneh:
@ManishMadan2882 the output.mp3 was something for me to check the streaming and test the api, you can run the elevenlabs.py separately to see that working. 

let me look into it tonight, now I am away from workspace

aalghooneh:
@ManishMadan2882 this works fine for me, did you set these variables in the code before trying?
self.api_key = 'ELEVENLABS_API_KEY'# here you should put your api key
self.model = "eleven_flash_v2_5"
self.voice = "VOICE_ID" # this is the hash code for the voice not the name!
self.write_audio = 1
the voice_id should be a hash, and elevenlabs_api_key should be a valid elevenlabs api

aalghooneh:
thanks @ManishMadan2882, 

it also needs authorization to deploy Vercel-docsgpt so it could merge.

aalghooneh:
@dartpain , @ManishMadan2882 , ?

PR #1593, opened 2025-01-20
style: use underscore for an unused loop variable

vil02:
This addresses the SC2034 warning.

Uses the underscore `_` instead of `i` in the _waiting for docker loop_ in `setup.sh`.

Because of _pure esthetics_. But it also resolved the [`SC2034`](https://www.shellcheck.net/wiki/SC2034) ShellCheck warning.

- **Other information**:
N/A


ManishMadan2882 (approved):
Good Catch @vil02  !

dartpain (approved):

PR #1597, opened 2025-01-22
Added Arabic language support

uka7:
Feature: This PR introduces Arabic language support.
Adding Arabic language support ensures that the application is accessible to Arabic-speaking users
- **Other information**:

### DIFF
--- a/frontend/src/locale/ar.json
+++ b/frontend/src/locale/ar.json
@@ -1,0 +1,191 @@
+{
+    "language": "العربية",
+    "chat": "دردشة",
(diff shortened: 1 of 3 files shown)


uka7:
@dartpain can you check this?

Charlesnorris509 (commented):
While im not familiar with arabic i think the intent behind this pull request is incredibly thoughfull, i will also attempt to complete a similar modification but for my own ethnic language which expand the scope of users/usage on the platform. Great JOB

PR #1609, opened 2025-01-30
Update agent.py

Charlesnorris509:
Add methods to handle new tool in `Agent` class

* **Agent class**:
  - Add `add_tool` method to load and configure new tools.
  - Add `handle_tool_action` method to execute actions for the new tool.


dartpain:
Really appreciate this PR, but agent is being completely changed now here:
https://github.com/arc53/DocsGPT/pull/1648
If you have any suggestion / fixes for it I would appreciate a lot!

PR #1628, opened 2025-02-11
Smoother transition in settings

ManishMadan2882:
     - The skeleton will be delayed for atleast 2 seconds to avoid flashing (`useLoaderState` custom hook)
     - color scheme and animations are uniform in all the loader components
     - Avoiding ternary operators for each loader variant, instead a separate renderMethod is mapped(ease of maintenance)
     - Advanced options drop gently in the Upload
     - Table and inline code blocks are improved in conversation bubble
   UI/UX


dartpain (approved):
All good, thank you!

PR #1680, opened 2025-03-04
fix: add missing env variable to docker compose for frontend

asminkarki012:
 bug fix
Fix Sources File Not Loading in Front End When EMBEDDINGS_NAME Environment Variable is Changed
[frontend/src/components/SourceDropdown.tsx](https://github.com/arc53/DocsGPT/blob/main/frontend/src/components/SourceDropdown.tsx)

- **Other information**:
None


asminkarki012:
Will be handled from backend side...

Issue #1699, opened 2025-03-14
🐛 Bug Report:  AzureOpenAILLM has incorrect signature,  Docker compose with deployment/docker-compose-azure.yaml fails


nickaggarwal:
### 📜 Description

When trying to start with azure_openai the deployment fails 

[2025-03-14 03:22:19,182] ERROR in routes: /api/answer - error: AzureOpenAILLM.__init__() missing 2 required positional arguments: 'openai_api_version' and 'deployment_name' - traceback: Traceback (most recent call last):
  File "/app/application/api/answer/routes.py", line 541, in post
    agent = AgentCreator.create_agent(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/application/agents/agent_creator.py", line 14, in create_agent
    return agent_class(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/app/application/agents/classic_agent.py", line 21, in __init__
    super().__init__(endpoint, llm_name, gpt_model, api_key, user_api_key)
  File "/app/application/agents/base.py", line 14, in __init__
    self.llm = LLMCreator.create_llm(
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/app/application/llm/llm_creator.py", line 32, in create_llm
    return llm_class(api_key, user_api_key, *args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: AzureOpenAILLM.__init__() missing 2 required positional arguments: 'openai_api_version' and 'deployment_name'

### 👟 Reproduction steps

Set the .env file with 

LLM_NAME=azure_openai
API_KEY=
MODEL_NAME=gpt-4o
OPENAI_API_KEY=
AZURE_EMBEDDINGS_DEPLOYMENT_NAME="text-embedding-3-large"
OPENAI_API_BASE=
OPENAI_API_VERSION="2024-11-20"
AZURE_DEPLOYMENT_NAME=gpt-4o

Run 

docker compose -f deployment/docker-compose-azure.yaml up -d

### 👍 Expected behavior

It should initialize the AzureOPENAI Class

Signature to be changed to 

def __init__(
        self, openai_api_key, openai_api_base,  *args, **kwargs
    ):

### 👎 Actual Behavior with Screenshots

[...]

Issue #1697, opened 2025-03-13
Something went wrong. Please check your internet connection.  / Ollama Setup on Ubuntu


DrNokkel:
I performed the standard installation using the setup.sh file.
After everything was successfully executed for the Ollama setup (without modifications, just the default settings), I can access the frontend. There, I can upload files and adjust options.
However, the chat does not work.
Keep saying this error message
![Image](https://github.com/user-attachments/assets/3cb144b0-6c5f-485a-b1e1-77e1cc70efb1)

Ollama is accessible via the default port and indicates that it is running.

I tried installing Ollama separately and linking it, but the same error occurs. In the terminal, I can chat with the LLM without any issues.

I also tried adjusting the .env file afterward, but that didn’t help either.

I am using the system with Ubuntu.

dartpain:
Hmm,

Can you provide me with your .env file and please tell me which ollama model you are using?

Also do you see any error inside your backend container?

DrNokkel:
My .env file 

API_KEY=None
LLM_NAME=openai
MODEL_NAME=llama3.2
VITE_API_STREAMING=true
OPENAI_BASE_URL=http://localhost:11434
EMBEDDINGS_NAME=huggingface_sentence-transformers/all-mpnet-base-v2

Model llama3.2 
![Image](https://github.com/user-attachments/assets/d2a40fab-bd18-4853-ba41-78447cecd79e)


......:~/DocsGPT$ curl http://localhost:11434/api/tags
{"models":[{"name":"llama3.2:latest","model":"llama3.2:latest","modified_at":"2025-03-13T14:54:04.558023555+01:00","size":2019393189,"digest":"a80c4f17acd55265feec403c7aef86be0c25983ab279d83f3bcd3abbcb5b8b72","details":{"parent_model":"","format":"gguf","family":"llama","families":["llama"],"parameter_size":"3.2B","quantization_level":"Q4_K_M"}}]}




...:/DocsGPT$ docker ps
CONTAINER ID   IMAGE                 COMMAND                  CREATED        STATUS             PORTS                      NAMES
12f24b40f91c   deployment-backend    "gunicorn -w 2 --tim…"   3 hours ago    Up About an hour   0.0.0.0:7091->7091/tcp     deployment-backend-1



Logs from backend

...:~/DocsGPT$ :~/DocsGdocker logs 12f24b40f91c
[2025-03-13 13:55:05 +0000] [1] [INFO] Starting gunicorn 23.0.0
[...]

Issue #1622, opened 2025-02-07
Something went wrong. Please check your internet connection.


amou1017:
Hello
I followed this document to install DocsGPT on ubuntu-24.04, but when asking the question I got the following error message 
"Something went wrong. Please check your internet connection."
The error message displayed on the terminal

backend_1   | [2025-02-07 08:39:04,785] ERROR in routes: /api/answer - error: No LLM class found for type  - traceback: Traceback (most recent call last):
backend_1   |   File "/app/application/api/answer/routes.py", line 495, in post
backend_1   |     for line in retriever.gen():
backend_1   |   File "/app/application/retriever/classic_rag.py", line 86, in gen
backend_1   |     agent = Agent(
backend_1   |             ^^^^^^
backend_1   |   File "/app/application/tools/agent.py", line 11, in __init__
backend_1   |     self.llm = LLMCreator.create_llm(
backend_1   |                ^^^^^^^^^^^^^^^^^^^^^^
backend_1   |   File "/app/application/llm/llm_creator.py", line 30, in create_llm
backend_1   |     raise ValueError(f"No LLM class found for type {type}")
backend_1   | ValueError: No LLM class found for type

Does anyone know the reason?
The attachment is my settings

amouc@docsgpt:~/DocsGPT$ vi .env

```
API_KEY=sk-proj-vEuP-xBgZaSYQf9kuyx3N5E73Z47XbPCbC9UQUo2D3EqaUahRvJ0UVNnc3XHt3NoT_9fsGEBXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXXX
LLM_NAME=openai
VITE_API_STREAMING=true
```

amouc@docsgpt:~/DocsGPT$ vi deployment/docker-compose.yaml

```

![Image](https://github.com/user-attachments/assets/e2caed9c-ecef-4ffa-a7a2-a1cdd133b552)
![Image](https://github.com/user-attachments/assets/fe4d7a88-975e-441a-805c-a7b04d5b56b4)

```

padma-ram:
I am facing the same issue.

ManishMadan2882:
Could you please share the issue in Discord https://discord.gg/5AqE3WkP ?

PR #1700, opened 2025-03-14
fix: docker compose up doesn't use the env 

ScriptScientist:
Bug fix
#1697
#1699
#1622  
- **Other information**:


ScriptScientist:
@dartpain I find it not use the generate env when start service by docker compose up, it only use on docker compose build

dartpain (approved):
Thanks!

PR #1742, opened 2025-04-15
File System Abstraction

ManishMadan2882:
      - Adding abstraction for uploading, retrieving, processing, deleting files 
      - Option for S3 and localStorage
      - Updated file attachment uploads 
       - ease of  swapping between storage(local/cloud)


Charlesnorris509 (changes requested):
Hey @manish It seems like you are done yet with this, but I came across some issues that I think you should take into consideration, The cuurent S3 connection script is implementing Boto3.client() but if you added a config section you would be able to ensure that transmission use SSL for security and in transit encryption. You should also try to overwrite the current boto3 retry components in the process_files() portion to limit unnecessary file downloads

dartpain:
Hey @Charlesnorris509 
I thought boto3 uses SSL by default.
CleanShot 2025-04-16 at 20 28 58@2x

Maybe im mistaken, can you provide some links to documentation about it?

Charlesnorris509:
@ManishMadan2882, the current version of Boto use SSL by default but since you've specified which version of the API you'll use, it might actually route it to an insecure endpoint.  Here's the link for the documentation https://boto3.amazonaws.com/v1/documentation/api/latest/guide/configuration.html

The proper Implementation should like this 'from botocore.config import Config

self.s3 = boto3.client(
    's3',
    aws_access_key_id=aws_access_key_id,
    aws_secret_access_key=aws_secret_access_key,
    config=Config(
        signature_version='s3v4',
        s3={'addressing_style': 'path'}
    )
)

ManishMadan2882:
@Charlesnorris509  ​In boto3 version 1.35.97, the default signature version used for signing AWS service requests is Signature Version 4 (s3v4).
https://boto3.amazonaws.com/v1/documentation/api/1.35.97/guide/configuration.html

Charlesnorris509:
Then sorry, i should have checked the version, and i think you should be fine then

dartpain (approved):

PR #1763, opened 2025-05-02
Always Render the Top Bar with Share (Mobile & Desktop)

aidanbennettjones:
- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...) Attempted Change To Always Render Top Bar With Share (on both Mobile and Desktop)

- **Other information**:


ManishMadan2882 (commented):
@aidanbennettjones Thanks for putting this together. The current implementation doesn't align with our design specs, so we'll be closing this PR

PR #1768, opened 2025-05-07
Attachments : File storage changes

ManishMadan2882:
     - In  `StoreAttachment` endpoint:
         - Upload the file to storage first
         - Get the metadata from the storage operation
         - Pass the path and metadata to the worker instead of file content
     - In  `attachment_worker`:
         - Remove the file_content from file_info
         - Get the file from storage using the path
         - Use the metadata from the upload operation
         - Process the file content as before
 avoids passing large binary data through Celery

- **Other information**:


dartpain (approved):
All good, thank you, lots of linting!

PR #1824, opened 2025-06-02
feat: reorganize LLM handler structure with better abstraction

### LINKED ISSUE #1769: 🚀 Feature: LLM handler Refactor
### 🔖 Feature description

- Simplify tool call logic and move it to LLM abstraction https://github.com/arc53/DocsGPT/blob/main/application/agents/llm_handler.py to llm
- Checks if specific LLM model support different feature (attachments, tools) 
- Rename env variable LLM_NAME into LLM_PROVIDER and MODEL_NAME -> LLM_NAME, MODEL_PATH -> LLM_PATH, MODEL_TOKEN_LIMITS -> LLM_TOKEN_LIMITS
- Include native support for fallback llm (FALLBACK_LLM_PROVIDE and FALLBACK_LLM_NAME) which are triggered on error for main LLM call.
- While streaming response, tools calls need to show up as they are being requested and when they are completed
- Adding agent implementations (types) should be much more simplified.

Go through the whole app and make sure that all names make sense in relation to LLM's,  models and so on

To clean up the code, Ensure consistent behavior

Change existing LLM abstraction.
Move logic from llm_handler in agents to llm abstraction

### 🔄️ Additional Information

_No response_

None

siiddhantt:
- resolves #1769 
-  ToolCall and LLMResponse dataclasses for standardization
- separate handler classes with common workflow for streaming and non-streaming llm responses
- fallback for llm requests


dartpain (approved):
All good, thank you!

PR #1838, opened 2025-06-11
Fixes ingestion of file with non-ascii characters in name

### LINKED ISSUE #1835: 🐛 Bug Report: Can't upload file with non-ASCII characters
### 📜 Description

If the uploaded filename contains any non-ASCII characters the app breaks. 

### 👟 Reproduction steps

1. Try to upload a source (potentially an attachment also). 
2. Choose any file which name contains contains non-ASCII characters (any non latin alphabet will work, feel free to use файл.pdf or 파일.pdf)

Specifying a separate name for the upload will also result in similar behaviour.  

### 👍 Expected behavior

We need to allow users to use non-ASCII characters in the filename or the name of the upload (`request.form["name"]`)
We also need to keep sanitizing filenames that will be used by the app, but will not be displayed to the user. 

The app stops working. 

### 💻 Operating system

MacOS

Chrome

Local dev server

_No response_

The error happens due to this import:
`from werkzeug.utils import secure_filename`
This `secure_filename` function completely removes all non-ASCII characters and sometimes even more. And this leads to no docs being provided later.
The problematic file is: `application/api/user/routes.py`. Specifically this route `@user_ns.route("/api/upload")`, but there might be more issues that I didn't find. 

I think that display source name needs to be separated from the system filename. 
We can utilise something like `from unidecode import unidecode` to prepare filename before sanitizing with `secure_filename` or generate something random (however this filename is displayed above chunks in the settings).
For example setting `filename = secure_filename(unidecode(file.filename))` allowed for a temporary fix, however the displayed source name is still incorrect.  
The displayed filename needs to be same as provided by the user for UX. 

### 📖 Relevant log output

```shell

```

None

### 🧑‍⚖️ Code of Conduct

ManishMadan2882:
  - added a utility method `safe_filename` which checks if the `secure_filename` method from werkzeug utility package returns an empty string
  - Generally the system would crash as this method would return "pdf"  for a file named файл.pdf or 파일.pdf
  - The `safe_filename` method handles this case, the string before the file extension if emtpy is substituted by a uuid, ensure that the name 
  - path of the original file, which is related to the storage is included in the mongodb `sources` document to keep the track of the user uploaded file
  - The UploadFile and StorageAttachments rely on `safe_filename`
   Closes #1835


dartpain:
Saw the PR,
Looks like new source has name with uuid in the UI.
UI should have original filename, but in the background we should be naming it with uuid

ManishMadan2882:
@dartpain Could you please review it again, the issue should be fixed.

Issue #1919, opened 2025-08-09
🐛 Bug Report: FeedBack Buttons Hover State


hanzalahwaheed:
### 📜 Description

The below recording illustrates that the hover state for the additional buttons is set wrong. The buttons in the response Bubble should be visible at all times (as it is in chatGPT).

https://github.com/user-attachments/assets/390b1ea4-ddb3-4860-9cc8-1c49fcf1a722

This looks out of place if clicked on like/dislike button as well.

<img width="870" height="521" alt="Image" src="https://github.com/user-attachments/assets/c16f3a4f-697d-4aa6-b518-b2978ce00f50" />

### 👟 Reproduction steps

1. Trigger a Response in Chat.
2. Click on the like button in the response bubble.
3. Move cursor out of the bubble.

### 👍 Expected behavior

The buttons in the response bubble should always be visible. Similar to chatGPT. Attaching screenshot below for reference:

<img width="829" height="236" alt="Image" src="https://github.com/user-attachments/assets/25092784-4f59-4b8c-a13e-cc074a344c0a" />

### 👎 Actual Behavior with Screenshots

<img width="870" height="521" alt="Image" src="https://github.com/user-attachments/assets/c16f3a4f-697d-4aa6-b518-b2978ce00f50" />

### 💻 Operating system

MacOS

### What browsers are you seeing the problem on?

Something else

### 🤖 What development environment are you experiencing this bug on?

Local dev server

### 🔒 Did you set the correct environment variables in the right path? List the environment variable names (not values please!)

Yes

### 📖 Relevant log output

```shell

```

### 👀 Have you spent some time to check if this bug has been raised before?

- [x] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

Yes, I am willing to submit a PR!

### 🧑‍⚖️ Code of Conduct

- [x] I agree to follow this project's Code of Conduct

ManishMadan2882:
@hanzalahwaheed Agree with you, the feedback options should be always accessible as demonstrated in the expected behaviour, Thanks, assigned to you!

PR #1920, opened 2025-08-11
fix: always show the response bubble buttons.

hanzalahwaheed:
This change introduces a fix to the buttons in the response bubbles of the chat UI. The buttons are always visible now and not just upon hovering. Replicated chatGPT like functionality.

https://github.com/arc53/DocsGPT/issues/1919

- **Other information**:

Some more suggestions can be taken up later:
1. Text font size for user and assistant message is different (14px and 16px respectively) in mobile view.
2. Copy btn should also have a hover effect similar to other buttons.

### DIFF
--- a/frontend/src/conversation/ConversationBubble.tsx
+++ b/frontend/src/conversation/ConversationBubble.tsx
@@ -565,16 +565,9 @@ const ConversationBubble = forwardRef<
                 <div>{retryBtn}</div>
               </div>
             )}
-            {handleFeedback && (
+            {handleFeedback && type !== 'ERROR' && (
               <>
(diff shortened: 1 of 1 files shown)


ManishMadan2882 (changes requested):
@hanzalahwaheed We need to show the like and dislike buttons, even when the feedback for one of the options is submitted, this is a bit different from the chatGPT example shared in the issue description but our application can handle to revert the user feedback
Image
Thanks!

hanzalahwaheed:
@ManishMadan2882 
done w the changes. plz check.

also fixed some redundancies

PR #1930, opened 2025-08-16
Added support for Multi select sources

### LINKED ISSUE #1912: 🚀 Feature: Multi select sources
### 🔖 Feature description

Make sure users can select and chat with multiple sources at a time

UX

Combine data for vectorsearch across multiple filters

### 🔄️ Additional Information

_No response_

None

Ankit-Matth:
Previously, the system only supported a single source at a time. This PR introduces support for **multi-select sources**, allowing users to select and chat with multiple document sources simultaneously.  

Fixes #1912 

- **Screen recording**:

https://github.com/user-attachments/assets/1579f966-e8c8-47b0-b672-8722f61338d6

### 🔍 Testing
- [x] Verified single-source selection still works as expected.  
- [x] Verified multiple sources can be selected simultaneously.  
- [x] Confirmed vector search results are aggregated correctly.


siiddhantt (commented):
LGTM!

Ankit-Matth:
> Hi @Ankit-Matth there are a few build errors so if you could fix those, and also we should be passing a list of sources only i.e no need to parse string and do it that way.

Got it 👍 Just to clarify, are you suggesting I should refactor the logic to directly pass a Python list of sources instead of the current comma-separated string parsing?

Also, where can I check the relevant build logs? I’d like to see the details of why the build failed...

siiddhantt:
> instead of the current comma-separated string parsing

Yes that would be good!

> where can I check the relevant build logs

You can just try to build locally using `npm run build` and see those errors during it.

siiddhantt:
Hi @Ankit-Matth looks good to me, thanks!

Just had one final suggestion, in the agents creation section we can choose a source too. So it would make sense to allow for multple source selection in there too, so if you're able to add that it'd be great else I can make the change for it.

Ankit-Matth:
@siiddhantt  Sure, I’ll add that tomorrow. I’m a bit tied up with some stuff today...

Ankit-Matth:
> Just had one final suggestion, in the agents creation section we can choose a source too. So it would make sense to allow for multiple source selection in there too, so if you're able to add that it'd be great else I can make the change for it.

Hey @siiddhantt, I’m down with a health issue, so I won’t be able to get to that change. If you can go ahead and make it, that’d be awesome. Thanks a lot!

Issue #1952, opened 2025-09-05
🚀 Feature: New Sources UI for Upload New functionality


dartpain:
### 🔖 Feature description

This feature introduces a new Sources UI that opens as a new modal when users click on the 'Upload New' button. 

<img width="822" height="808" alt="Image" src="https://github.com/user-attachments/assets/db0bd47f-27d3-456a-9513-12929ee526e3" />

Also minor edits:
- Make sure that the list of available sources depend on configuration. Basically dont show google drive if oath is not configured
- Also add connector type in the DB

### 🎤 Why is this feature needed ?

Better UX

### ✌️ How do you aim to achieve this?

**Design Reference:** https://www.figma.com/design/OXLtrl1EAy885to6S69554/DocsGPT?node-id=0-1&p=f

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### Are you willing to submit PR?

None

Ankit-Matth:
Can I work on it?

dartpain:
@ManishMadan2882 is already assigned for it.

PR #1954, opened 2025-09-08
Refinements: Files uploaded via Connectors

ManishMadan2882:
* What kind of change does this PR introduce?
    - Added Google File picker and `drive.file` scope for the drive uploads
    - Files uploaded via Connectors like drive are stored as `type : connector:file` in sources collection
    - Separated FilePicker(for remote connectors only), GoogleDrivePicker form Upload.tsx; adding new types for the form fields - `remote_file_picker`, `local_file_picker`, `google_drive_picker`
    - Added search functionality with debouncing to the FilePicker, with the updated UI 
    - Confirm sync form the user, in the Sources when the source type is `connector:*`
    - table UI for the file picker
    - refactored Upload - minimised complexity related to types in the ingestor.ts, combined IngestorFormSchemas with the available options with optional validation(if validation fails would not render the option, renders by default)
* Why was this change needed?
    - TypeScript noise. Removing them keeps the component focused and reduces maintenance.(Upload.tsx and Ingestor.ts)
    - Minimising the Google OAuth scope to `drive.file`(Non-sensitive)
    - UI update for the FilePicker component

* Other information
Add these environment variables from your Google console for testing
The API key is from the Google Picker API, make sure to set the javascript origins for the client as frontend url
      - `VITE_GOOGLE_API_KEY`
      - `VITE_GOOGLE_CLIENT_ID`


ManishMadan2882:
Closes #1952

PR #1956, opened 2025-09-10
refactor: replace Math.random with crypto.randomUUID for diagram IDs

jayamrutiya:
### Description
Replaced `Math.random()` with `crypto.randomUUID()` when generating diagramId for Mermaid diagrams.

### Motivation
- `Math.random()` is not cryptographically secure and was flagged by Sonar rule `TypeScript:S2245`.
- Although the ID is only used for UI elements, using `crypto.randomUUID()` ensures uniqueness without triggering security warnings.
- No external dependencies are required since crypto is a built-in API in modern browsers and Node.js.

### Changes
- Updated diagramId generation in MermaidRenderer to use `crypto.randomUUID()`.

### Benefits
- Eliminates Sonar security warning (S2245).
- More reliable and future-proof unique ID generation.
- Cleaner and safer implementation without additional dependencies.

### DIFF
--- a/frontend/src/components/MermaidRenderer.tsx
+++ b/frontend/src/components/MermaidRenderer.tsx
@@ -17,7 +17,7 @@ const MermaidRenderer: React.FC<MermaidRendererProps> = ({
 }) => {
   const [isDarkTheme] = useDarkTheme();
   const diagramId = useRef(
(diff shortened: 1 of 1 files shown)


dartpain:
Its does not need to be cryptographically secure, its just for UI to not confuse elements.

PR #1979, opened 2025-09-25
Fix #162: Add throttling to frontend API calls

### LINKED ISSUE #162: Add throttling to frontend api calls

Shlokyy4:
This PR addresses issue #162 by adding throttling to the post method in frontend/src/api/client.ts to prevent excessive API calls (e.g., rapid chat submissions). The change uses lodash.throttle to limit requests to a 500ms interval, ensuring only the first call in that window proceeds.

Key Changes:

- Added import { throttle } from 'lodash'; to enable throttling.
- Implemented a throttledPost function with a 500ms delay, wrapped around the original fetch logic.
- Updated apiClient.post to use throttledPost, applying the throttling.

Commit Details:

- Files changed: frontend/src/api/client.ts, frontend/package.json, frontend/package-lock.json.
- Commit message: Fix #162: Add throttling to frontend API calls.

Testing:

- Restarted DocsGPT using .\setup.ps1 (option 1 for public API).
- Verified at http://localhost:5173 by spamming chat inputs, confirming ~500ms spacing in the Network tab (F12).

### DIFF
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -1,3 +1,5 @@
+import { throttle } from 'lodash';
+
 export const baseURL =
   import.meta.env.VITE_API_HOST || 'https://docsapi.arc53.com';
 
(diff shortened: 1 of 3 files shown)


ManishMadan2882:
@Shlokyy4 The implementation does not match our requirements, throttling should not be applied globally to the routes

Issue #2016, opened 2025-10-05
UI/UX Design Fix Implementation (Mobile)


poncema4:
![image](https://github.com/user-attachments/assets/419c319c-d7d1-49c1-a237-000c0a3f3bc8)

When trying to share the link it is blocked off and not able to see the full message to the user. Should be a quick fix:

Solutions:

Fix the dimensions of the feedback for sharing the link

Extend or shorten the message for users on a mobile device

P.S

Can make a PR once I implemented the changes and thoroughly test these changes if fixed

Pratheek2474:
Hi, i want to work on this.. can you assign it to me?

anshumaaaan:
Hey! I've submitted a PR to fix that layout overflow issue. Let me know if there's anything else needed.

poncema4:
Thank you @ManishMadan2882, solved.

ManishMadan2882:
@anshumaaaan @poncema4 
A review of the architecture indicates that the changes cannot be applied to the shared `WrapperModal` component, as it would incorrectly affect other components that reuse it.

Please move your changes directly into `frontend/src/modals/ShareConversationModal.tsx`.

Since the PRs were identical, would prefer based on First-Come, First-Served (FCFS)

Thanks!

anshumaaaan:
Hey @ManishMadan2882 I have made the required changes to frontend/src/modals/ShareConversationModal.tsx.
Please check and let me know if there's anything else needed.

PR #2018, opened 2025-10-05
Improve ShareConversationModal layout and spacing

anshumaaaan:
Issue:
#2016 
Bug fix
The sharing button in WrapperModal was going out of bounds and users could not see the full message when trying to share a link.
This change fixes the modal dimensions to properly display the feedback for sharing links, ensuring better usability on both desktop and mobile devices.
- **Other information**:
Updated WrapperModal.tsx to set a maximum height (max-h-[80vh]) and width (w-[600px], max-w-[90vw]) for the modal and enabled vertical scrolling if content overflows.

Before and after:
image
image


ManishMadan2882 (changes requested):
The changes cannot be in the generic WrapperModal because it is reused. The fix must be moved to the specific component: frontend/src/modals/ShareConversationModal.tsx.
https://github.com/arc53/DocsGPT/issues/2016#issuecomment-3375424701

PR #2024, opened 2025-10-06
fix the command for installing on windows

kartik912:

Docs update

---

The command for installing on Windows in the README.md file was incorrect for some environments. The `.` in `./setup.ps1` could cause path resolution issues in certain PowerShell or command-prompt configurations. 

This change corrects the command from `PowerShell -ExecutionPolicy Bypass -File ./setup.ps1` to `PowerShell -ExecutionPolicy Bypass -File setup.ps1` to ensure it runs reliably for all Windows users.

---
- **Other information**:

This is a minor documentation fix to improve the setup experience on Windows. No functional code has been changed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,7 +115,7 @@ A more detailed [Quickstart](https://docs.docsgpt.cloud/quickstart) is available
 2. **Run the PowerShell setup script:**
 
    ```powershell
-   PowerShell -ExecutionPolicy Bypass -File .\setup.ps1
+   PowerShell -ExecutionPolicy Bypass -File setup.ps1
    ```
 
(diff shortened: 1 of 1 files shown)


dartpain:
The .\ prefix is actually PowerShell best practice

kartik912:
> The .\ prefix is actually PowerShell best practice

@dartpain , i am fully aware about this practice , but when i run it on my windows, it was not working, so that is why i suggested this change.

dartpain:
So it breaks on windows if you are using the "./" thing?

dartpain:
ill have to double check it on my windows system later today too

Issue #2034, opened 2025-10-07
UI: Center Text by removing hidden div margins


hanzalahwaheed:
<img width="860" height="206" alt="Image" src="https://github.com/user-attachments/assets/ff2eec6a-79ee-4c8d-a4cc-3ff29deb2ba4" />

<img width="574" height="259" alt="Image" src="https://github.com/user-attachments/assets/4e77b591-9743-4136-9bbf-da7c5dc222af" />

In the above image, the title for the custom agent isn't centered cause we are still rendering a hidden div whose margin is taking up space. 

Can be fixed by rendering conditionally.

Happy to create a PR

Hrithik-12:
Can I work on this issue please !!

hanzalahwaheed:
@dartpain PR ready here: https://github.com/arc53/DocsGPT/pull/2040

also, idk if such a small PR would get accepted in hf25, but plz add the necessary labels - `hacktoberfest`, `hacktoberfest-accepted`

PR #2040, opened 2025-10-07
fix(ui): centered title due to improper obj check

hanzalahwaheed:
issue: https://github.com/arc53/DocsGPT/issues/2034

The bug was that an empty obj evaluates to true, hence rendering the empty div and its css.

- **Other information**:

perfectly centered, as all things should be :P
image


ManishMadan2882 (approved):
Good catch, Thanks !

Issue #1984, opened 2025-09-26
🚀 Feature: Agent API security improvements


pabik:
### 🔖 Feature description

This feature proposes the implementation of critical security and quality-of-life enhancements for the public-facing Agent API. As agents are designed to be embedded and shared, it's essential to provide users with robust tools to control access and prevent abuse.

I propose adding two key features to the agent configuration settings:
1. CORS Origin Whitelisting: An option for users to specify a whitelist of allowed domains. This will ensure that the agent's API can only be called from trusted websites. 
2. API Limits: A mechanism to set limits on token usage (or maybe number of requests). This is crucial for protecting against denial-of-service (DoS) attacks and managing the costs associated with LLM calls.

### 🎤 Why is this feature needed ?

Currently, once an agent is made public, its API endpoint is open to the world. This poses significant risks for some users. 

### ✌️ How do you aim to achieve this?

The implementation can be broken down into two parts, targeting the backend API and the frontend Agent Advanced section.

1. CORS Origin Whitelisting:

Frontend: 
- In the Agent Advanced section, add a new text input field labeled "Allowed Origins". This field should accept a comma-separated list of URLs.

Backend:
- Store the list of allowed origins in the database, associated with the specific agent.
- In the API middleware that handles agent requests, check the incoming request's Origin header if enabled.

2. API Limits

Frontend: 
- In the Agent Advanced section, add new text input field labeled "Usage limits"

Backend:
- In the API middleware that handles agent requests, check usage before processing if enabled.

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### Are you willing to submit PR?

None

ardafincan:
Hey @pabik , I want to implement this. I will handle the backend part first but I am not sure about the frontend part. I will try that too after backend. Can you assign me this?

dartpain:
Thank you, @ardafincan 
Assigned to you.
In terms of frontend changes, try to hide the settings inside the advanced section please. If anything make a mock desing we will ask our UI designer to adjust it a bit later on.

ravi-hash:
Can you please assign to me

ardafincan:
Hey, I have already implemented the api limiting part, the only remaining is whitelisting

ardafincan:
I had some confusion about the whitelisting part, but I guess I got it right now :D. I have such a squeezed schedule but I will try to implement it in 3-4 days max.

dartpain:
[...]

PR #2041, opened 2025-10-08
Feat: Agent Token or Request Limiting

ardafincan:
### ###
This PR introduces daily token usage and request limiting for agent api's. User can specify token usage or request limits while creating an agent.

As mentioned at #1984 this feature improves api security by limiting request or token usage per 24 hours.

- **Other information**:
Original issue #1984 includes also a CORS origin whitelisting but I decided to convert that to another PR because there are some issues with that feature to be discussed. I will be writing about that under the related issue.


ardafincan:
@dartpain yeah you are right, those slipped out of my sight. They are fixed right now.

Issue #2046, opened 2025-10-09
Refactoring The Notification Section


omaremadcc:
Refactoring the notification to: 
 - check if this message is sent to the user before
 - Switch to use "sonner" to handle notification logic efficiently
 - Use gradient instead of images to improve loading time
 - Move all the logic to "NotificationToast.tsx" to keep "App.tsx" clean and minimal

URAYUSHJAIN:
@OM3X4 i can try this

omaremadcc:
I have made it myself> @OM3X4 i can try this

ManishMadan2882:
I don’t think the changes here address the issue we’re looking to solve. I would suggest looking at the already opened issues or please make sure to wait for assignment /confirmation before starting work next time.

Issue #2023, opened 2025-10-06
🚀 Feature: Notification section


dartpain:
### 🔖 Feature description

New clickable notification section with a clickable link.
Should be responsive and scalable
Also should be set by 2 env variables (one for text, 2nd for link)

<img width="1540" height="936" alt="Image" src="https://github.com/user-attachments/assets/033b8a20-58eb-47a1-9355-b6a0d6935ecf" />

### 🎤 Why is this feature needed ?

better UX

### ✌️ How do you aim to achieve this?

Frontend changes

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### Are you willing to submit PR?

None

Mariam-Saeed:
Hello, I'd like to work on this one.

kartik912:
hello,
i am looking for my first hacktoberfest contribution.
please it would be great if you can assign me this

shinigami-777:
Hi, Can I work on this?

Deekshit014:
Hello! I would love to work on this issue (#2023). I have a local environment running.
Could you please assign it to me? Thank you for your consideration!

omaremadcc:
Hi maintainers , I have implemented the desired feature in pr #2045 , It is exactly follow the figma design , You can change both the message and link for .env , the message only show once because it check the local storage

PR #2047, opened 2025-10-09
Refactoring the notification Section

omaremadcc:
Refactoring the notification Section to:
    check if this message is sent to the user before
    Switch to use "sonner" to handle notification logic efficiently
    Use gradient instead of images to improve loading time
    Move all the logic to "NotificationToast.tsx" to keep "App.tsx" clean and minimal

- **What kind of change does this PR introduce?** (Bug fix, feature, docs update, ...) : Refactoring

- **Why was this change needed?** (You can also link to an open issue here): issue #2046 

- **Other information**:

### DIFF
--- a/frontend/.env.development
+++ b/frontend/.env.development
@@ -2,5 +2,6 @@
 VITE_BASE_URL=http://localhost:5173
 VITE_API_HOST=http://127.0.0.1:7091
 VITE_API_STREAMING=true
+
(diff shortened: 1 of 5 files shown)


ManishMadan2882:
Hi @OM3X4 
These changes are not exactly same as the Figma design shared in original issue https://github.com/arc53/DocsGPT/issues/2023
main priority is that the UI remains consistent with the design

PR #2066, opened 2025-10-14
Refactor: use async/await in fetchChunks for correct error handling (typescript:S4822)

jayamrutiya:
### Description

This PR refactors the `fetchChunks` function to comply with **Sonar rule `typescript:S4822`**, which warns when a `try...catch` is used around asynchronous code without an `await`.  

The previous implementation used `.then()` and `.catch()` promise chaining within a `try` block — meaning async errors were never caught by the outer `catch`.  

This update converts the function to `async/await` syntax for proper error handling and improves code clarity.

### Changes Made
- Converted `fetchChunks` to an `async` function.  
- Replaced `.then()` and `.catch()` chains with `await` and `try...catch`.  
- Added a `finally` block to ensure `setLoading(false)` runs on both success and error.  
- Improved readability and reduced duplicate loading state updates.

### Benefits:

- Proper async error handling (resolves Sonar rule typescript:S4822)
- Cleaner and more readable code
- Consistent loading state management
- Prevents unhandled promise rejections

### Refrence:
Sonar Rule: typescript:S4822: https://rules.sonarsource.com/typescript/RSPEC-4822/
 – "Consider using 'await' for the promise inside this 'try' or replace it with 'Promise.prototype.catch(...)' usage"

### DIFF
--- a/frontend/src/components/Chunks.tsx
+++ b/frontend/src/components/Chunks.tsx
@@ -136,35 +136,37 @@ const Chunks: React.FC<ChunksProps> = ({
 
   const pathParts = path ? path.split('/') : [];
 
-  const fetchChunks = () => {
-    setLoading(true);
-    try {
-      userService
-        .getDocumentChunks(documentId, page, perPage, token, path, searchTerm)
-        .then((response) => {
-          if (!response.ok) {
-            setLoading(false);
-            setPaginatedChunks([]);
-            throw new Error('Failed to fetch chunks data');
-          }
-          return response.json();
-        })
-        .then((data) => {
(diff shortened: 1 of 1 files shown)


jayamrutiya:
@ManishMadan2882 
Makes sense 👍 — I went with the async/await approach mainly for better readability, but I’m fine simplifying it by just removing the redundant try/catch if we prefer a minimal change here.

Let me know if we just want the minimal change or if it’s okay to keep the async/await version.

ManishMadan2882 (approved):
Thanks for the update and for addressing the feedback. I’m fine with the async/await as well !

Issue #2094, opened 2025-10-23
🐛 Bug Report: Delete conversations modal 'X' button broken


hanzalahwaheed:
### 📜 Description

minor bug reg delete all modal

### 👟 Reproduction steps

Settings -> General -> Delete All -> Try to close modal by `X` button

### 👍 Expected behavior

Clicking on X should close the modal.

### 👎 Actual Behavior with Screenshots

<img width="541" height="180" alt="Image" src="https://github.com/user-attachments/assets/2b0eb173-8ff8-4839-837e-9b1ed84e585b" />

Clicking on 'X' does nothing.

### 💻 Operating system

MacOS

### What browsers are you seeing the problem on?

Chrome

### 🤖 What development environment are you experiencing this bug on?

Local dev server

### 📖 Relevant log output

```shell

```

### 👀 Have you spent some time to check if this bug has been raised before?

- [x] I checked and didn't find similar issue

### 🔗 Are you willing to submit PR?

Yes, I am willing to submit a PR!

### 🧑‍⚖️ Code of Conduct

- [x] I agree to follow this project's Code of Conduct

hanzalahwaheed:
Please assign :)

hanzalahwaheed:
PR here: https://github.com/arc53/DocsGPT/pull/2109

PR #2109, opened 2025-10-27
Fix/x btn

hanzalahwaheed:
Fixes a minor bug that a certain modal X btn was broken.

https://github.com/arc53/DocsGPT/issues/2094

- **Other information**:
There is a slight redundancy with `handleCancel` and `handleClose` now as both do the same thing. Please let me know if this works.


ManishMadan2882:
@hanzalahwaheed The issue was recently resolved on the main branch, sorry for the confusion as your issue was opened before the fix was merged
Thanks!

Issue #2177, opened 2025-12-03
🚀 Feature: Autoselect input text box on pageload


dartpain:
### 🔖 Feature description

Basically autoselects text input box such that user can start typing right away

### 🎤 Why is this feature needed ?

Better UX

### ✌️ How do you aim to achieve this?

Frontent changes

### 👀 Have you spent some time to check if this feature request has been raised before?

- [x] I checked and didn't find similar issue

### Are you willing to submit PR?

None

ZainabTravadi:
Hi! I'd like to take this issue.

My approach will be to add a `ref` to the chat input component and use a `useEffect` hook to automatically focus the input on page load or when the chat view mounts. This keeps the change fully frontend and avoids any impact on existing functionality.

If this approach works for you, please assign the issue to me.

PR #2180, opened 2025-12-05
feat(frontend): add reliable autofocus to chat message textarea and input component

ZainabTravadi:
### Summary

This PR implements the feature requested in issue **#2177** by adding reliable autofocus behavior to the chat message input field. When the page loads or the chat UI mounts, users can immediately begin typing without clicking the input box—improving UX and aligning with expected chat application behavior.

### What was happening before

* The chat textarea (`MessageInput.tsx`) did not consistently receive focus on initial load.
* React re-renders triggered by conversation fetches or state updates caused the textarea to lose focus.
* The existing `autoFocus` prop was defined but never applied reliably to the actual textarea element.

### What this PR changes

* Applies autofocus to the main chat `<textarea>` using a delayed focus mechanism inside `useEffect` to ensure focus persists across initial renders.
* Supports the existing `autoFocus` prop (default: `true`).
* Adds stable autofocus handling to the custom `<Input>` component as well (via ref + effect).
* Ensures users can begin typing immediately after loading the chat page.

### Why this approach

React components inside DocsGPT’s chat UI re-render during initialization (loading conversations, state hydration). Immediate focusing inside `useEffect` can fire too early. Using a delayed focus resolves the race condition and ensures consistent user experience.

### Testing

* Verified that typing works immediately after page load with zero user interaction.
* Cursor blinks in the message textarea on refresh.
* Autofocus persists even during UI re-renders.
* No backend or breaking changes introduced.

### Result

A smoother, more intuitive chat experience: users can start typing right away, without needing to click the input field.


siiddhantt:
Issue #2177 closed with PR #2194

PR #2259, opened 2026-01-09
Refactor: Extract agent configuration logic from StreamProcessor

AkashBhadana:
-  What kind of change does this PR introduce?
- Code Refactoring / Quality Improvement

- Why was this change needed?
- The StreamProcessor class was handling too many different tasks (configuration, data loading, etc.), making it difficult to maintain and test.
- This change makes the code cleaner by moving all the agent configuration logic into a new, dedicated AgentConfigurator class.
- It's the first step in making the code easier to understand and improve in the future.

- Other Information
- This is a pure refactoring, so there are no changes to the application's functionality.

### DIFF
--- a/application/api/answer/services/agent_configurator.py
+++ b/application/api/answer/services/agent_configurator.py
@@ -1,0 +1,278 @@
+import datetime
+import logging
+from typing import Any, Dict, Optional
+
+from bson.dbref import DBRef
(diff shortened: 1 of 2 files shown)


ManishMadan2882:
the changes incorporate unintended behaviour.

PR #2268, opened 2026-01-27
fix(frontend): fix the input styling on renaming chat

IRjSI:
- Bug fix (frontend UI)

This change aligns the input styling with the rest of the conversation tile and improves usability.

- **Other information**:
No functional behavior was changed; this PR only affects styling during the chat rename state.


ManishMadan2882:
I see that it’s marked as a bug fix for the frontend UI and that it updates the input styling during the chat rename state. Could you clarify what the specific bug or user issue was that prompted this change?

IRjSI:
The input area on renaming a chat was not styled, that is not rounded and the a blue outline. I fixed the corners (rounded) and disabled outline.

PR #2273, opened 2026-02-04
add navbar toggling shortcut

IRjSI:
Feature: Added a keyboard shortcut to toggle the navbar for improved usability.

To allow quicker access to the navbar without relying on mouse interaction.

- **Other information:**
Shortcut: Ctrl + Shift + S

### DIFF
--- a/frontend/package-lock.json
+++ b/frontend/package-lock.json
@@ -27,6 +27,7 @@
         "react-redux": "^9.2.0",
         "react-router-dom": "^7.6.1",
         "react-syntax-highlighter": "^15.6.1",
+        "react-tooltip": "^5.30.0",
         "rehype-katex": "^7.0.1",
         "remark-gfm": "^4.0.0",
         "remark-math": "^6.0.0",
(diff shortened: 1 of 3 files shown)


aaron-seq:
## Code Review

**Disclosure:** This review was conducted with AI assistance and has been fully reviewed and validated.

### Overview
The implementation adds a keyboard shortcut (Ctrl+Shift+S) to toggle the navbar, which improves usability. The code structure is clean with proper event listener cleanup.

### Issues Identified:

1. **Missing dependency in useEffect** (Line 300): The dependency array is empty `[]`, but `setNavOpen` is used inside the effect. While React guarantees `setNavOpen` stability, it's better practice to include it or use a comment explaining why it's omitted.

2. **Potential keyboard shortcut conflict**: Ctrl+Shift+S might conflict with browser's "Save As" on some platforms. Consider using a less common combination or document this known limitation.

3. **Lack of user-facing documentation**: The shortcut is not visible in the UI. Users won't know about this feature unless they read the code. Consider adding a tooltip or keyboard shortcuts help panel.

### Positive Aspects:

- Proper cleanup function in useEffect
- Cross-platform support (ctrlKey || metaKey)
- preventDefault() correctly prevents browser default behavior
- Event listener properly removed on unmount

### CI/CD Status:
All checks passed. The Bandit Security Scan warning is about CodeQL Action deprecation (not related to this PR).

### Recommendation:
The code logic is correct and functional. However, I recommend addressing the dependency array and considering documentation before merging.

IRjSI:
Fixed the issues.

Add dependency in useEffect.
No conflict with "save as" as the default behaviour is prevented.
Add a tooltip, so the users will now know about the shortcut feature.

Issue #2278, opened 2026-02-14
Fix XSS vulnerability: replace dangerouslySetInnerHTML with safe React rendering in PromptsModal


yusufazam225:
The PromptTextarea component in PromptsModal.tsx used dangerouslySetInnerHTML to render syntax-highlighted prompt text. While a custom escapeHtml function was applied, this approach is fragile -- any missed escape path (from future edits, shared prompts, or encoding edge cases) could allow script injection. The fix replaces raw HTML string injection with native React element rendering, which auto-escapes all content by default, making XSS structurally impossible.

ManishMadan2882:
@yusufazam225 
For the current PromptsModal, the vulnerability is not practically exploitable because the escapeHtml function correctly escapes user-controlled content before rendering.

However, relying on manual escaping is fragile, good catch! The real risk appears if the textarea input is later rendered using innerHTML in a different context without proper sanitization.

PR #2279, opened 2026-02-14
Fix XSS vulnerability: replace dangerouslySetInnerHTML with safe React rendering in PromptsModal

yusufazam225:
#2278 

what i did:-
Replaced dangerouslySetInnerHTML with React children rendering in PromptTextarea. Instead of building a raw HTML string and injecting it, highlightPromptVariables now returns an array of React elements. This lets React auto-escape all text content, removing the XSS risk. The escapeHtml helper was also removed since it's no longer needed.


ManishMadan2882 (approved):
Good catch @yusufazam225 !

PR #2283, opened 2026-02-19
Sharepoint

ManishMadan2882:
     - OneDrive - add integration to existing connector types
     - User can access files from their drive and the files shared to them
     - The /drive/sharedWithMe endpoint in Microsoft Graph is currently deprecated and degraded, and in practice it often returns only a limited subset of shared items. At the moment, Microsoft does not provide a clearly documented replacement endpoint for reliably enumerating files shared with a user.
     To work around this limitation, this implementation uses the Graph Search API with KQL to retrieve shared documents. However, the search API can sometimes return duplicate or unexpected hits, so the implementation performs a broader search and then deduplicates results by drive item ID before returning them.
     This approach works for work/school accounts (Entra ID tenants) but unfortunately does not apply to personal Microsoft accounts (MSA) due to platform restrictions.
     Users can use their private and shared files(entra id accounts) from OneDrive as knowledge bases/source
    

- **Other information**:
image

### DIFF
--- a/tests/parser/remote/test_share_point_loader.py
+++ b/tests/parser/remote/test_share_point_loader.py
@@ -67,8 +67,9 @@ class TestSharePointLoaderLoadFileById:
 
     @patch("application.parser.connectors.share_point.loader.requests.get")
     @patch("application.parser.connectors.share_point.loader.SharePointAuth.get_token_info_from_session")
+    @patch("application.parser.connectors.share_point.loader.SharePointAuth.__init__", return_value=None)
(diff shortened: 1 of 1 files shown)


ManishMadan2882:
@siiddhantt Thanks for the review, have addressed the concerns.
.env-template is for backend configuration. `VITE_SHARE_POINT_CLIENT_ID ` is a frontend specific env var

siiddhantt:
Thanks @ManishMadan2882 will do one final integration test today and let you know!

PR #2331, opened 2026-03-26
security: migrate AES-CBC to AES-GCM authenticated encryption

### LINKED ISSUE #2503: 🐛 Bug Report: Stored credential encryption lacks integrity protection
### 📜 Description

The pre-PR implementation of `application/security/encryption.py` encrypted stored credentials with AES-CBC and custom padding, but did not authenticate the ciphertext, IV, salt, or user context with an integrity tag.

As a result, a party that can modify a stored encrypted credential blob can make controlled changes to decrypted plaintext without the application detecting tampering. This can affect API tokens and connector credentials stored through this helper.

PR #2331 migrates new credential encryption to AES-GCM and binds `user_id` as AAD. The PR branch also keeps a legacy CBC fallback for backward compatibility, so existing CBC ciphertexts should be rotated or re-encrypted after deployment.

Affected component:

- `application/security/encryption.py`
- `encrypt_credentials(credentials: dict, user_id: str)`
- `decrypt_credentials(encrypted_data: str, user_id: str)`

### 👟 Reproduction steps

1. Check out the vulnerable baseline before PR #2331, for example local `main` at commit `06e827573cac6ad3b445ffea0fb08169071c1b50`.
2. Set only the test environment variable name `ENCRYPTION_SECRET_KEY` to a controlled test value.
3. Encrypt a synthetic credential object such as `{"a": "AAAA", "b": 1}` for `user-123`.
4. Base64-decode the stored blob.
5. Flip byte 7 of the IV with `ord("A") ^ ord("Z")`.
6. Base64-encode the modified blob and call `decrypt_credentials(..., "user-123")`.
7. Observe that the decrypted object changes to `{"a": "ZAAA", "b": 1}` instead of failing authentication.
8. Check out PR #2331 head `5ca9a64e9eae6a7423b667bdeae9386200355478`.
9. Repeat tampering against a newly encrypted GCM blob.
10. Observe that the decrypt path returns `{}` for tampered GCM ciphertext and `{}` for cross-user replay.

Local validation command used:

```shell
PYTHONPATH=/home/ubuntu/summer/agentscan-repos/repos/222 /tmp/docsgpt-pr2331-venv/bin/python -m pytest tests/security/test_encryption.py -q -o addopts=''
```

Result:

```shell
11 passed in 0.56s
```

Controlled PoC result:

```shell
legacy_pre_pr_targeted_tamper_result= {'a': 'ZAAA', 'b': 1}
fixed_gcm_tamper_result= {}
fixed_gcm_cross_user_replay_result= {}
fixed_format_version_byte= 1

### 👍 Expected behavior

Credential ciphertext should be authenticated. Any modification to the salt, nonce/IV, ciphertext, tag, or user binding should fail decryption and return no credential data.

Newly encrypted credentials should use an AEAD mode such as AES-GCM. Legacy CBC credentials should be migrated or rotated so the stored data is no longer accepted without an integrity check.

In the vulnerable baseline, AES-CBC decryption accepts modified ciphertext when the resulting plaintext is syntactically valid JSON. There is no authentication tag, so targeted IV manipulation can alter plaintext bytes in the first decrypted block without detection.

No screenshots are applicable because this is a backend cryptographic integrity issue.


### 💻 Operating system

Linux

Something else

Local dev server

Only environment variable names used during validation:

- `ENCRYPTION_SECRET_KEY`

No secret values are included in this report.

PR #2331:

- `https://github.com/arc53/DocsGPT/pull/2331`

PR metadata observed during validation:

- Title: `security: migrate AES-CBC to AES-GCM authenticated encryption`
- State: `open`
- Head branch: `fix/aes-gcm-authenticated-encryption`
- Head commit: `5ca9a64e9eae6a7423b667bdeae9386200355478`
- Modified files:
  - `application/security/encryption.py`
  - `tests/security/test_encryption.py`

Security policy note: `SECURITY.md` asks vulnerability reporters to email `security@arc53.com`. This file is prepared in the repository's bug report format, but the project security policy should be followed before public disclosure.


### 📖 Relevant log output

```shell
legacy_pre_pr_targeted_tamper_result= {'a': 'ZAAA', 'b': 1}
fixed_gcm_tamper_result= {}
fixed_gcm_cross_user_replay_result= {}
fixed_format_version_byte= 1
```

### 🧑‍⚖️ Code of Conduct

3em0:
## Summary
- **Replaced AES-CBC (no message authentication) with AES-GCM** in `application/security/encryption.py`, providing both confidentiality and integrity for stored credentials
- **Backward-compatible decryption**: existing CBC-encrypted data is auto-detected via a version byte prefix and decrypted through the legacy path
- **Added tests** for GCM round-trip, legacy CBC compatibility, tamper detection, and cross-user replay rejection

Fixes #2503

## Motivation
AES-CBC without HMAC lacks ciphertext integrity protection. While the practical Padding Oracle attack surface is limited (no direct decrypt endpoint), the absence of authenticated encryption is a cryptographic deficiency. AES-GCM resolves this in a single primitive.

## Changes
| File | Change |
|------|--------|
| `application/security/encryption.py` | `encrypt_credentials` -> AES-256-GCM with `user_id` bound as AAD; `decrypt_credentials` -> GCM-first versioned decrypt with legacy CBC fallback |
| `tests/security/test_encryption.py` | Updated round-trip test; added legacy CBC compatibility, tamper detection, cross-user replay, version-byte collision fallback, and corrupt legacy payload coverage |

## Test plan
- [x] `/tmp/docsgpt-pr2331-venv/bin/python -m pytest tests/security/test_encryption.py -q --no-cov` - 11 passed
- [x] GCM encrypt/decrypt round-trip verified
- [x] Legacy CBC data still decrypts correctly
- [x] Tampered GCM ciphertext correctly rejected (returns `{}`)
- [x] GCM ciphertext encrypted for one user is rejected for another user

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/application/security/encryption.py
+++ b/application/security/encryption.py
@@ -2,6 +2,7 @@ import base64
 import json
 import os
 
+from cryptography.exceptions import InvalidTag
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.ciphers import algorithms, Cipher, modes
@@ -37,10 +38,11 @@ def encrypt_credentials(credentials: dict, user_id: str) -> str:
         nonce = os.urandom(12)
         key = _derive_key(user_id, salt)
 
+        aad = user_id.encode()
         json_bytes = json.dumps(credentials).encode()
 
         aesgcm = AESGCM(key)
-        ciphertext_with_tag = aesgcm.encrypt(nonce, json_bytes, None)
+        ciphertext_with_tag = aesgcm.encrypt(nonce, json_bytes, aad)
 
         result = _VERSION_GCM + salt + nonce + ciphertext_with_tag
         return base64.b64encode(result).decode()
@@ -55,10 +57,16 @@ def decrypt_credentials(encrypted_data: str, user_id: str) -> dict:
     try:
         data = base64.b64decode(encrypted_data.encode())
 
+        # Try GCM first (version-prefixed format). If the GCM tag check fails
+        # (InvalidTag), fall back to legacy CBC — this avoids the 1/256 collision
+        # where a legacy salt's first byte happens to be 0x01.
         if data[0:1] == _VERSION_GCM:
-            return _decrypt_gcm(data[1:], user_id)
-        else:
-            return _decrypt_legacy_cbc(data, user_id)
+            try:
+                return _decrypt_gcm(data[1:], user_id)
+            except InvalidTag:
+                pass
+
+        return _decrypt_legacy_cbc(data, user_id)
     except Exception as e:
         print(f"Warning: Failed to decrypt credentials: {e}")
         return {}
@@ -70,9 +78,10 @@ def _decrypt_gcm(data: bytes, user_id: str) -> dict:
     ciphertext_with_tag = data[28:]
 
     key = _derive_key(user_id, salt)
+    aad = user_id.encode()
 
     aesgcm = AESGCM(key)
-    plaintext = aesgcm.decrypt(nonce, ciphertext_with_tag, None)
+    plaintext = aesgcm.decrypt(nonce, ciphertext_with_tag, aad)
 
     return json.loads(plaintext.decode())
 
--- a/application/security/encryption.py
+++ b/application/security/encryption.py
@@ -5,9 +5,12 @@ import os
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes
 from cryptography.hazmat.primitives.ciphers import algorithms, Cipher, modes
+from cryptography.hazmat.primitives.ciphers.aead import AESGCM
 from cryptography.hazmat.primitives.kdf.pbkdf2 import PBKDF2HMAC
 
 from application.core.settings import settings
+
+_VERSION_GCM = b"\x01"
 
 
 def _derive_key(user_id: str, salt: bytes) -> bytes:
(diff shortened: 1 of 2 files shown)


3em0:
@larinam  @larinam @sfx101 hello?

JiwaniZakir (commented):
The version-byte detection in `decrypt_credentials` has a subtle edge case: legacy CBC blobs use a random salt as their first 16 bytes, so roughly 1 in 256 existing CBC-encrypted credentials will have a first byte of `\x01` and be silently misrouted to `_decrypt_gcm`, causing authentication to fail and returning `{}` instead of the correct data. A safer approach would be to attempt GCM first and fall back to CBC only on `InvalidTag`, or to use a longer/more distinctive version prefix that can't collide with valid salt bytes.

Additionally, both `_decrypt_gcm` and `encrypt_credentials` pass `None` as the AAD parameter to `AESGCM`. Binding the authentication tag to something like the `salt` or `user_id` bytes would prevent a scenario where a ciphertext blob is transplanted between users (if key derivation were weakened in the future) — the GCM tag would cover only the ciphertext, not its intended owner. The `test_tampered_gcm_ciphertext_returns_empty` test verifies bit-flipping is caught but doesn't test cross-user replay.

One missing test scenario: `_decrypt_legacy_cbc` is never tested with an invalid/truncated blob on its own — the existing `test_decrypt_credentials_returns_empty_for_invalid_input` uses a random string rather than a structurally valid-but-corrupt CBC payload, so the `_unpad_data` error path in the legacy branch isn't explicitly exercised.

3em0:
did it！

PR #2423, opened 2026-04-22
feat: improve conversation scrolling and loading states effectively

siiddhantt:
- Added meta tags for mobile web app capabilities in index.html.
- Introduced loading state management for agents in Navigation component.
- Updated NewAgent component to handle agent chunks as strings.
- Enhanced Conversation component to display loading indicators for agents.
- Improved ConversationBubble styles and interactions for better UX.
- Adjusted ConversationMessages to manage scrolling and loading states effectively.
- Refactored useMediaQuery hook for better performance and responsiveness.
- Updated CSS for consistent styling and animations.
- Extended preferenceSlice to include agent loading state.
- Initialized agent loading state in the Redux store.

### DIFF
--- a/frontend/src/conversation/ConversationMessages.tsx
+++ b/frontend/src/conversation/ConversationMessages.tsx
@@ -75,6 +75,7 @@ export default function ConversationMessages({
   const scrollAnimFrameRef = useRef<number | null>(null);
   const settleAnimFrameRef = useRef<number | null>(null);
   const showButtonTimerRef = useRef<ReturnType<typeof setTimeout>>(undefined);
+  const shadowRafRef = useRef<number | null>(null);
 
   useEffect(() => {
     return () => {
@@ -194,7 +195,13 @@ export default function ConversationMessages({
     if (!el) return;
     const atBottom = isAtBottom();
     const dist = Math.max(0, el.scrollHeight - el.scrollTop - el.clientHeight);
-    onScrolledFromBottom?.(dist);
+    if (onScrolledFromBottom) {
+      if (shadowRafRef.current !== null) cancelAnimationFrame(shadowRafRef.current);
+      shadowRafRef.current = requestAnimationFrame(() => {
+        onScrolledFromBottom(dist);
+        shadowRafRef.current = null;
+      });
+    }
     if (atBottom) {
       isAutoScrollingRef.current = false;
       if (userInterruptedRef.current && wasNotAtBottomRef.current) {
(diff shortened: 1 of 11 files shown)


frontend/src/conversation/ConversationMessages.tsx
```diff
@@ -193,45 +170,83 @@ export default function ConversationMessages({
...
     }
 
     const isNewMessage = queries.length > prevQueries.length;
     prevQueriesRef.current = queries;
 
-    scrollConversationToBottom(isNewMessage ? false : true);
+    if (isNewMessage) {
+      userInterruptedRef.current = false;
+      setInterrupted(false);
+      scrollToBottom(true);
```
ManishMadan2882: changes the auto-scroll on a new message from smooth to instant, which I believe is a regression from the previous UX.
could you confirm if this is intentional?

frontend/src/conversation/ConversationMessages.tsx
```diff
@@ -129,61 +133,34 @@ export default function ConversationMessages({
...
-              setButtonHidden();
-            }
-            isAutoScrollingRef.current = false;
-          }, 500);
-        }
-      });
-    },
-    [isAtBottom, setButtonHidden],
-  );
+  const scrollToBottom = useCallback((instant = true) => {
```
ManishMadan2882: changes the intended UX, the `scrollToBottom` is turned instant by default.

ManishMadan2882 (changes requested):
Hey @siiddhantt noticed few issues in the scrolling mechanism which potentially break the intended UX for the scroll.

 Regression: autoscroll vs. user scroll "tug-of-war" during streaming

reintroduces the behavior that was intentionally fixed in #420e9d3[ ((feat) conversation: scroll experience)](https://github.com/arc53/DocsGPT/pull/2374). When the user scrolls upward while a response is streaming, the view snaps back to the bottom instead of respecting the user's intent.

Repro
- Send a prompt that yields a long streamed response.
- While it's streaming and the view is pinned to the bottom, wheel/touch upward by a small amount (a few px).
- Observe: the viewport is pulled back down on the next streamed chunk. Only scrolls larger than  SCROLL_THRESHOLD (10px) in a single event survive.

frontend/src/conversation/ConversationMessages.tsx
siiddhantt: there was an issue with the mermaid component being rendered while streaming of the answers that got fixed thru this, I've made some changes now to improve the scroll and keep it working without breaking with mermaid

siiddhantt:
@ManishMadan2882 if you could check out the new changes and how the scroll feels now, do lmk thanks!
- added a scroll based opacity to the shadow above input box so the transition on when the shadow is shown is even smoother
- the scrolling can be interrupted during streaming of answer, smoother scrolling without breaking during rendering of mermaid components
- fixed sources sidebar and other smaller changes

PR #2449, opened 2026-05-03
feat: cancel previous chat API calls using AbortController

### LINKED ISSUE #162: Add throttling to frontend api calls

SHRAVANIRANE:
What kind of change does this PR introduce?
Feature / Performance improvement

Why was this change needed?
When users send messages rapidly, multiple API requests are triggered for chat endpoints. This can increase backend load and lead to outdated responses being processed.
This change ensures that only the latest request is handled by cancelling previous in-flight requests using AbortController.
Fixes #162

Other information
- Scoped to /chat and /stream endpoints to avoid impacting other APIs
- Uses AbortController instead of debouncing for real-time request handling
- Includes safe cleanup logic to prevent race conditions
- Improves responsiveness and reduces redundant API calls

Debouncing delays requests, whereas chat systems benefit more from cancelling stale requests. This approach ensures only the most recent user input is processed.

### DIFF
--- a/frontend/src/api/client.ts
+++ b/frontend/src/api/client.ts
@@ -18,36 +18,77 @@ const getHeaders = (
   return headers;
 };
 
+// Replaces the debounce map. Stores active request controllers by URL.
+const abortControllersMap: Map<string, AbortController> = new Map();
+
+// Helper to catch 4xx/5xx HTTP errors and throw them properly
+const handleResponse = (response: Response): Response => {
+  if (!response.ok) {
+    throw new Error(`HTTP Error: ${response.status} ${response.statusText}`);
+  }
+  return response;
+};
+
 const apiClient = {
   get: (
     url: string,
     token: string | null,
     headers = {},
     signal?: AbortSignal,
-  ): Promise<any> =>
+  ): Promise<Response> =>
     fetch(`${baseURL}${url}`, {
       method: 'GET',
       headers: getHeaders(token, headers),
       signal,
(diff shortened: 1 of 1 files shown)


SHRAVANIRANE:
Hi, I’ve implemented request cancellation for chat endpoints using AbortController.  
Please let me know if any changes or improvements are needed.

tejinderpa:
Hey @SHRAVANIRANE,

I noticed the recent PR implementing request cancellation using `AbortController` for chat endpoints — that makes sense for real-time interactions 

My initial approach was slightly different: I was planning to introduce a more general-purpose `ThrottleManager` that would handle:

-> request throttling/debouncing
-> in-flight request deduplication across components
-> flexible keying (including non-serializable inputs)

This would be applied first to search flows (e.g. `Sources.tsx`) and potentially extended elsewhere.

Given the new PR, would you prefer:

1. Keeping the AbortController approach scoped to chat, and layering a throttling/deduplication utility on top for other APIs, or
2. Consolidating both ideas into a shared request management layer?

Also, for AbortSignal coordination in deduplicated requests, I was considering a reference-counting approach (only abort when all consumers release). Happy to adjust based on your preferred pattern.

Let me know how you'd like me to proceed — I can align my implementation accordingly.

ManishMadan2882:
This PR only does per-URL cancellation for two endpoints, and only in post. Throttling/dedupe across the rest of the surface is untouched.

PR #2459, opened 2026-05-08
Add authentication and per-user authorization to /api/images endpoint

sebastionoss:
## Summary

The `GET /api/images/<path:image_path>` endpoint in `application/api/user/attachments/routes.py` (the `ServeImage` resource) serves files from the configured upload storage with no authentication and no per-user authorization. The only check on the pre-fix path was `".." in image_path or image_path.startswith("/") or "\x00" in image_path`, which provides shallow CWE-22 hardening but does nothing about *who* is allowed to read a given file.

Because uploaded attachments are stored under `<UPLOAD_FOLDER>/<user>/attachments/<attachment_id>/<filename>` (see line 164 of the same module), any unauthenticated network client that learns or guesses such a path — for example via a leaked attachment ID, predictable filename, indexable storage backend, or a referer/log leak — can fetch another user's uploaded image. This is a missing-authentication / cross-tenant access issue (CWE-306 / CWE-639) layered on top of weak path validation (CWE-22).

- **Affected file/route:** `application/api/user/attachments/routes.py`, `ServeImage.get` → `GET /api/images/<path:image_path>`
- **CWE:** primarily CWE-306 (Missing Authentication) and CWE-639 (Authorization Bypass Through User-Controlled Key), with CWE-22 hardening
- **Severity:** High — unauthenticated cross-tenant read of user-uploaded files

### Data flow

1. `image_path` arrives directly from the URL with no auth gate (the global `@app.before_request authenticate_request` only *populates* `request.decoded_token`; individual routes must opt in, and this one didn't).
2. After a substring `..` check, the raw path is passed to `storage.get_file(image_path)`.
3. The storage backend resolves and returns the file bytes — to anyone who asked.

## Fix

Three changes to `ServeImage.get`, mirroring the pattern already used by the other endpoints in this module (which all call `_resolve_authenticated_user()` first):

1. **Authentication** — call the existing `_resolve_authenticated_user()` helper. If it returns a Flask error response, propagate it; if it returns no user, return `401 Authentication required`.
2. **Path validation** — normalize via `posixpath.normpath(image_path)` and reject paths that start with `..`, start with `/`, or contain a `..` segment. This is more robust than the previous substring `".." in image_path` test (which is bypassable in some encodings and inconsistent with the storage layer's own resolution) and matches what `posixpath` will actually produce.
3. **Authorization** — require the normalized path to start with `<UPLOAD_FOLDER>/<auth_user>/`, so an authenticated user can only fetch files under their own user prefix. Return `403 Forbidden` otherwise.

The downstream `storage.get_file` is now called with the normalized path, so any path canonicalization the storage layer performs sees the same value we authorized against.

## Tests

Added `tests/test_serve_image_auth.py` (218 lines) covering:

- Unauthenticated requests are rejected with 401.
- Authenticated requests for another user's path are rejected with 403.
- Authenticated requests for the user's own path succeed and return the expected content.
- Path traversal attempts (`..`, leading `/`, embedded `..` segments, NUL byte) are rejected with 400.
- API-key authentication path (matching the helper's existing behavior) is exercised.

The diff is minimal: 38 lines changed in `routes.py` and a new test file. No other modules touched.

## Security analysis

The exploit preconditions are simply (a) network reachability to `/api/images/` and (b) knowledge or a guess of a victim path component such as an attachment ID. Attachment IDs are MongoDB `ObjectId`s and filenames are user-supplied — neither is a security boundary, and IDs leak through normal product surfaces (chat history exports, logs, embeds). After the fix, an attacker without a valid session/API key gets 401, and an authenticated attacker attempting to read another tenant's file gets 403 before any storage call is made.

Before submitting, we tried to disprove this. We checked whether the global `authenticate_request` `before_request` hook would block unauthenticated callers — it does not; it only decodes a token if present and lets the request continue, leaving auth enforcement to each route. We checked whether `LocalStorage._get_full_path` would catch cross-user access via realpath containment — it only ensures the path stays within the storage base directory, not within the caller's user subdirectory, so it does not prevent cross-tenant reads. We also confirmed the substring `..` check was the only validation and that other endpoints in the same file already follow the `_resolve_authenticated_user()` + user-prefix pattern this PR adopts, so the change is consistent with existing conventions rather than introducing a new one.

cc @lewiswigmore

### DIFF
--- a/application/api/user/attachments/routes.py
+++ b/application/api/user/attachments/routes.py
@@ -1,6 +1,7 @@
 """File attachments and media routes."""
 
 import os
+import posixpath
 import tempfile
 from pathlib import Path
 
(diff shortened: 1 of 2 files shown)


andesyteoss:

PR #2460, opened 2026-05-08
security: enforce JWT expiration in simple_jwt and session_jwt auth modes

sebastionoss:
## Summary

JWTs issued by DocsGPT in `simple_jwt` and `session_jwt` auth modes never expire, and the server explicitly disables expiration verification when decoding them. Any token that leaks (logs, browser storage via XSS, MITM on a misconfigured deployment, shared device, accidental commit, etc.) grants permanent access to the API for as long as `JWT_SECRET_KEY` remains unchanged.

- **CWE-613**: Insufficient Session Expiration
- **CWE-287**: Improper Authentication
- **Affected files**: `application/app.py`, `application/auth.py`
- **Severity**: High when `AUTH_TYPE` is `simple_jwt` or `session_jwt`

## Vulnerability details

Two issues combine to make tokens permanently valid:

1. **Tokens are minted without an `exp` claim.**

   In `application/app.py`, the startup token for `simple_jwt`:
   ```python
   payload = {"sub": "local"}
   SIMPLE_JWT_TOKEN = jwt.encode(payload, settings.JWT_SECRET_KEY, algorithm="HS256")
   ```
   And the per-session token for `session_jwt` in `/api/generate_token`:
   ```python
   token = jwt.encode({"sub": new_user_id}, settings.JWT_SECRET_KEY, algorithm="HS256")
   ```
   Neither payload contains `exp` or `iat`.

2. **The decoder explicitly disables expiration checking.**

   In `application/auth.py`'s `handle_auth`:
   ```python
   decoded_token = jwt.decode(
       jwt_token,
       settings.JWT_SECRET_KEY,
       algorithms=["HS256"],
       options={"verify_exp": False},
   )
   ```

So even if a token did carry `exp`, the server would ignore it. `handle_auth` is the only JWT-decoding path in the application — there is no alternate validator that would catch this.

## Fix

- Add a small helper `generate_token_claims()` in `application/auth.py` that returns `{iat, exp}` with a 24-hour lifetime.
- Merge those claims into both token payloads at issuance (`simple_jwt` startup token and the `session_jwt` `/api/generate_token` route).
- Remove `options={"verify_exp": False}` from `jwt.decode` so `python-jose` enforces the `exp` claim.

The 24-hour default is conservative and matches common JWT session practice; it can be made configurable in a follow-up if desired. For `simple_jwt` deployments the operational implication is that the startup token must be regenerated (server restart, or a future enhancement to refresh it) every 24 hours — this is a usability trade-off, not a security regression, and is the standard expectation for bearer tokens.

## Tests

- Updated `tests/test_auth.py` to match the new `jwt.decode` call (no `verify_exp` option).
- Added `tests/test_cwe287_jwt_expiry.py` covering:
  - Issued tokens contain `iat` and `exp` claims with the expected lifetime.
  - Expired tokens are rejected by `handle_auth`.
  - Valid (unexpired) tokens are still accepted.
  - Both `simple_jwt` and `session_jwt` issuance paths produce expiring tokens.

All auth tests pass locally.

## Why this is exploitable

Preconditions are minimal:
- The deployment uses `AUTH_TYPE=simple_jwt` or `AUTH_TYPE=session_jwt` (both are first-class supported modes).
- The attacker obtains one valid JWT — possible via a number of routine paths: the startup token printed to logs in `simple_jwt` mode, browser storage exposure, network capture against a non-TLS or misconfigured proxy deployment, an accidental copy-paste into a chat/issue, or a shared workstation.

Once obtained, the token is a permanent credential. There is no rotation, no revocation list, and no expiry — the only way to invalidate it is to rotate `JWT_SECRET_KEY`, which invalidates every other user's token at the same time.

## Adversarial review

Before submitting, we tried to disprove this. We checked whether any other layer (reverse proxy assumptions, a second decode path, or framework defaults) might already enforce expiry — `handle_auth` is the only decoder, `verify_exp: False` is explicit, and `python-jose` does not expire tokens that lack an `exp` claim regardless of options. We also considered whether the startup-printed `simple_jwt` token is "operator-only" and therefore not really a credential leak risk; in practice it ends up in deployment logs, container stdout, and orchestrator dashboards, which is exactly the kind of secondary surface that token expiration is designed to limit. The fix is a strict improvement with no behavioural change for tokens used within their lifetime.

cc @lewiswigmore

### DIFF
--- a/application/app.py
+++ b/application/app.py
@@ -6,7 +6,7 @@ import dotenv
 from flask import Flask, jsonify, redirect, request
 from jose import jwt
 
-from application.auth import handle_auth
+from application.auth import generate_token_claims, handle_auth
 
 from application.core.logging_config import setup_logging
 
@@ -62,6 +62,7 @@ if settings.AUTH_TYPE in ("simple_jwt", "session_jwt") and not settings.JWT_SECR
 SIMPLE_JWT_TOKEN = None
 if settings.AUTH_TYPE == "simple_jwt":
     payload = {"sub": "local"}
+    payload.update(generate_token_claims())
     SIMPLE_JWT_TOKEN = jwt.encode(payload, settings.JWT_SECRET_KEY, algorithm="HS256")
     print(f"Generated Simple JWT Token: {SIMPLE_JWT_TOKEN}")
 
@@ -92,8 +93,10 @@ def get_config():
 def generate_token():
     if settings.AUTH_TYPE == "session_jwt":
         new_user_id = str(uuid.uuid4())
+        payload = {"sub": new_user_id}
+        payload.update(generate_token_claims())
         token = jwt.encode(
-            {"sub": new_user_id}, settings.JWT_SECRET_KEY, algorithm="HS256"
+            payload, settings.JWT_SECRET_KEY, algorithm="HS256"
         )
         return jsonify({"token": token})
     return jsonify({"error": "Token generation not allowed in current auth mode"}), 400
--- a/application/auth.py
+++ b/application/auth.py
@@ -1,6 +1,20 @@
+from datetime import datetime, timedelta, timezone
+
 from jose import jwt
 
 from application.core.settings import settings
+
+# Default token lifetime: 24 hours
+JWT_TOKEN_LIFETIME_HOURS = 24
+
+
+def generate_token_claims():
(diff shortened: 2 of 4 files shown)


andesyteoss:

PR #2466, opened 2026-05-10
Harden agent API tool against SSRF via redirects and path-param injection (CWE-918)

andesyteoss:
## Summary

This PR hardens `APITool._make_api_call()` (`application/agents/tools/api_tool.py`) against two SSRF gaps (CWE-918) that bypass the existing `validate_url()` allowlist check.

**1. Redirect-based SSRF bypass.** All seven HTTP method branches (`GET`, `POST`, `PUT`, `DELETE`, `PATCH`, `HEAD`, `OPTIONS`) call `requests.<method>(...)` without `allow_redirects=False`. `requests` follows 3xx redirects by default. `validate_url()` is only applied to the initial URL; a legitimate, allow-listed external endpoint can return `Location: http://169.254.169.254/...` (or any RFC1918 / loopback target) and the redirected request is sent without re-validation. The fix passes `allow_redirects=False` on every branch.

**2. Path-parameter injection.** When the configured URL template contains `{param}` placeholders, the substitution at line 91-95 inserts the raw `str(query_params[param_name])` into the URL. Path parameter values originate from LLM tool-call arguments and are influenced by retrieved document content / user queries (a realistic prompt-injection surface in an agent system). A value containing `/`, `?`, `#`, or `@` can change the request path, append/override query parameters on the configured host, or alter the URL fragment. The fix wraps the value in `urllib.parse.quote(value, safe="")` before substitution so reserved characters are percent-encoded.

## Fix

`application/agents/tools/api_tool.py`:
- Import `quote` alongside `urlencode`.
- Encode path-param values with `quote(str(value), safe="")` before `request_url.replace(...)`.
- Add `allow_redirects=False` to all `requests.{get,post,put,delete,patch,head,options}` calls.

The diff is small and surgical (~30 lines of production code).

## Tests

Added `tests/agents/tools/test_cwe918_api_tool_ssrf.py` (220 lines) covering:
- `allow_redirects=False` is set on each of the 7 HTTP methods.
- A redirect to `169.254.169.254` (AWS metadata) is not followed.
- Path values containing `/`, `?`, `#`, `@` are percent-encoded and do not alter URL structure.
- Normal alphanumeric path params still work end-to-end.

Existing api_tool tests still pass; behaviour for non-malicious inputs is unchanged.

## Adversarial review

Before submitting we tried to disprove this. The pre-existing `validate_url()` checks both the configured URL and the post-substitution URL, which initially looked like sufficient defence. It isn't:
- Redirect bypass: `validate_url()` runs *before* the HTTP call, never on the `Location` header. The default `allow_redirects=True` means the second hop is unchecked. This is exploitable whenever an admin configures an api_tool pointing at any host that an attacker can influence to issue a redirect (their own server, a vulnerable third-party, or an open-redirect endpoint on a trusted host).
- Path-param bypass: the post-substitution `validate_url()` re-check does catch host-changing payloads like `{id}` → `evil.com/x` because `urlparse` would surface a different `netloc`. It does **not** catch payloads that keep the host but mutate the path/query/fragment — for example a value `123?admin=true` that flips a privileged query flag, or `..%2fadmin` that traverses to a different endpoint on the same allow-listed host. Encoding is the correct defence-in-depth.

Preconditions are realistic for DocsGPT's agent architecture (a configured api_tool plus prompt-injection or attacker-controlled redirect endpoint) and do not by themselves grant equivalent SSRF capability.

## Notes for maintainers

- I noticed `SECURITY.md` prefers GitHub's private vulnerability reporting flow. I don't have permission to open an advisory on this repo, so I'm submitting publicly. Happy to redact and move to a private advisory if you'd prefer — just let me know.
- No behavioural change for legitimate API tool usage: redirects are uncommon in REST API integrations, and percent-encoding is what callers normally want for path params.

### DIFF
--- a/application/agents/tools/api_tool.py
+++ b/application/agents/tools/api_tool.py
@@ -2,7 +2,7 @@ import json
 import logging
 import re
 from typing import Any, Dict, Optional
-from urllib.parse import urlencode
+from urllib.parse import quote, urlencode
 
 import requests
 
@@ -91,8 +91,14 @@ class APITool(Tool):
                 for match in re.finditer(r"\{([^}]+)\}", request_url):
                     param_name = match.group(1)
                     if param_name in query_params:
+                        # URL-encode path parameter values to prevent path
+                        # traversal and query/fragment injection via
+                        # LLM-controlled inputs.
+                        safe_value = quote(
+                            str(query_params[param_name]), safe=""
+                        )
                         request_url = request_url.replace(
-                            f"{{{param_name}}}", str(query_params[param_name])
+                            f"{{{param_name}}}", safe_value
                         )
                         path_params_used.add(param_name)
             remaining_params = {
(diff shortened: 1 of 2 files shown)


andesyteoss:

PR #2565, opened 2026-06-24
fix: replace print() with logger in encryption.py and classic_rag.py

### LINKED ISSUE #2561: fix: print() statements in encryption.py and classic_rag.py bypass structured logger
## Bug Description

Three `print()` calls bypass the application's structured logging system, risking exposure of rephrased user query text and credential error details to unmanaged stdout.

| File | Line | Content |
|------|------|---------|
| `application/retriever/classic_rag.py` | 141 | `print(f"Rephrased query: {rephrased_query}")` |
| `application/security/encryption.py` | 48 | `print(f"Warning: Failed to encrypt credentials: {e}")` |
| `application/security/encryption.py` | 72 | `print(f"Warning: Failed to decrypt credentials: {e}")` |

`encryption.py` has no `logging` import at all. `classic_rag.py` uses `logging.error()` on the very next line (144) but `print()` on line 141.

## Expected Behavior

All runtime output should go through Python's `logging` module so log-level filtering, structured log collectors, and redaction pipelines work correctly.

## Proposed Fix

Replace `print()` with `logger.debug()` / `logger.warning()` and add a module-level `logger = logging.getLogger(__name__)` to both files.

immuhammadfurqan:
## Summary

- Replaces 3 bare `print()` calls with `logger.debug()` / `logger.warning()` using the standard `logging` module
- Adds `import logging` + `logger = logging.getLogger(__name__)` to `encryption.py` which had none
- Adds `logger = logging.getLogger(__name__)` to `classic_rag.py` and migrates all bare `logging.xxx()` calls to use it

## Motivation

Print statements bypass log-level filtering and structured logging pipelines. The encryption failure messages and rephrased query text could be exposed in production stdout without redaction.

Closes #2561

## Test Plan
- [x] Trigger a query rephrase and verify debug log appears at DEBUG level, not stdout
- [x] Trigger an encryption/decryption error and verify warning appears in logs, not stdout
- [x] Confirm no `print()` calls remain in the modified files

### DIFF
--- a/application/retriever/classic_rag.py
+++ b/application/retriever/classic_rag.py
@@ -7,6 +7,8 @@ from application.retriever.labels import labels_from_metadata
 from application.utils import num_tokens_from_string
 from application.vectorstore.vector_creator import VectorCreator
 
+logger = logging.getLogger(__name__)
+
 
 class ClassicRAG(BaseRetriever):
     def __init__(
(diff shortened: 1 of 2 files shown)


ManishMadan2882 (approved):
There was a ruff lint issue, I pushed a commit. seems reasonable overall. thanks!

PR #2574, opened 2026-06-26
fix: enable Chokidar polling for Vite HMR in Docker on Windows

### LINKED ISSUE #2456: 🐛 Bug Report: Vite's HMR update is not working in docker compose on windows
### 📜 Description

When running the frontend with docker-compose-local.yaml on Windows, changes inside src are reflected in the mounted file system, but Vite does not restart or hot-reload. The dev server stays running and only picks up changes after a manual container restart

This is happening because the file system change events are not getting delivered from Windows to the Linux container
to Chokidar ( vite relies on chokidar for hmr update )

### 👟 Reproduction steps

1. compose up using docker-compose-local.yaml
2. Change something in the frontend src folder
3. Observe logs of Docker Compose 

### 👍 Expected behavior

Editing any file in the src should trigger Vite HMR immediately
No manual container restart should be required

https://github.com/user-attachments/assets/f509061c-ce39-42b7-b8a5-f503fa8a07c2

This video shows that the changes are reflected in the mounted volume, but the Vite server is getting file change events, resulting in no HMR updates

### 💻 Operating system

Windows

Chrome

Docker

_No response_

_No response_

### 📖 Relevant log output

```shell

```

### 🧑‍⚖️ Code of Conduct

mmustafasenoglu:
## Summary

Fixes #2456

Native fs events do not propagate from Windows hosts into Linux containers, causing Vite's HMR to stop detecting file changes when using .

## Changes

- ****: When the  environment variable is set, Chokidar falls back to polling with a 300ms interval. This reliably detects file changes across the Docker volume mount boundary on Windows.
- ****: Added  to the frontend service environment.

## How it works

Vite uses Chokidar for file watching. On Linux, Chokidar uses  which works natively. However, when running Docker on Windows, the Linux container cannot receive filesystem events from the Windows host through the mounted volume. Polling bypasses this limitation by periodically checking file timestamps.

The  option is only enabled when  is set, so native Linux/Mac development environments are not affected by the additional CPU overhead.

## Testing

1. Run 
2. Edit any file in 
3. Vite HMR should now detect the change and hot-reload automatically

### DIFF
--- a/deployment/docker-compose-local.yaml
+++ b/deployment/docker-compose-local.yaml
@@ -7,6 +7,7 @@ services:
       - VITE_API_HOST=http://localhost:7091
       - VITE_API_STREAMING=$VITE_API_STREAMING
       - VITE_EMBEDDINGS_NAME=$EMBEDDINGS_NAME
+      - DOCKER=true
     ports:
       - "5173:5173"
 
--- a/frontend/vite.config.ts
+++ b/frontend/vite.config.ts
@@ -22,6 +22,12 @@ export default defineConfig(({ mode }) => {
             .map((h) => h.trim())
             .filter(Boolean)
         : [],
+      // Use polling for file watching when running inside Docker.
+      // Native fs events do not propagate from Windows hosts into Linux
+      // containers, so Chokidar falls back to polling which works reliably.
+      watch: env.DOCKER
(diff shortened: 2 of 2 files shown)


ManishMadan2882 (approved):
Works fine, tested on Windows + WSL + Docker. Thanks!

PR #2575, opened 2026-06-26
feat: optimize frontend Dockerfile with multi-stage build

### LINKED ISSUE #2452: 🚀 Feature: Dockerfile is too heavy in frontend, Reduce the image size
### 🔖 Feature description

The current Dockerfile of the frontend is unoptimized
It takes around 1.11 GB now

I locally optimized it and reduced its size by 850MB
It now only takes 267 MB

Image

Reducing the size of the image helps in both development and deployment 
It will reduce the dev setup time

By using multistage builds and removing unnecessary things from the Dockerfile

### 🔄️ Additional Information

_No response_

mmustafasenoglu:
## Summary

Fixes #2452

Reduces the frontend Docker image size from ~1.1GB to ~267MB (76% reduction) by using multi-stage builds and a lightweight base image.

### Dockerfile ()
- **Base image**:  →  (significantly smaller)
- **Multi-stage build**: Separate stages for development, build, and production
- **Package install**:  →  for reproducible builds
- **Production**: Uses  to serve static files

### New files
- ****: Nginx configuration for SPA routing and asset caching
- ****: Excludes , ,  from build context

### Docker Compose files
- : Added 
- : Added 
- : Added 

## Build targets

| Target | Used by | Description |
|--------|---------|-------------|
|  | docker-compose-local.yaml, docker-compose.yaml | Runs Vite dev server with HMR |
|  | docker-compose-azure.yaml | Builds static files and serves via nginx |

## Backward compatibility

- Production still serves on port 5173 to match existing setup
- Development target works exactly as before (same Vite dev server)
- No environment variable changes required

### DIFF
--- a/deployment/docker-compose-azure.yaml
+++ b/deployment/docker-compose-azure.yaml
@@ -1,6 +1,8 @@
 services:
   frontend:
-    build: ../frontend
+    build:
+      context: ../frontend
+      target: production
     environment:
       - VITE_API_HOST=http://localhost:7091
       - VITE_API_STREAMING=$VITE_API_STREAMING
--- a/deployment/docker-compose-local.yaml
+++ b/deployment/docker-compose-local.yaml
@@ -1,6 +1,8 @@
 services:
   frontend:
-    build: ../frontend
+    build:
+      context: ../frontend
+      target: development
     volumes:
     - ../frontend/src:/app/src
     environment:
--- a/deployment/docker-compose.yaml
+++ b/deployment/docker-compose.yaml
@@ -1,7 +1,9 @@
 name: docsgpt-oss
 services:
   frontend:
-    build: ../frontend
+    build:
+      context: ../frontend
+      target: development
     volumes:
       - ../frontend/src:/app/src
     environment:
--- a/frontend/.dockerignore
+++ b/frontend/.dockerignore
@@ -1,0 +1,5 @@
+node_modules
+dist
+.git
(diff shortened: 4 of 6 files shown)


ManishMadan2882:
Thanks for the contribution, @mmustafasenoglu 

I'm going to close this as a duplicate of #2453, which targets the same issue (#2452) . The two are near-identical: same 6 files, a byte-for-byte identical nginx.conf and .dockerignore, and the same node:22-alpine --> build --> nginx:alpine Dockerfile structure with the same compose target wiring. #2453 already has review history and iteration, so we'll consolidate there to avoid splitting the effort.

Worth flagging that both PRs share the same blocking gap, so closing this isn't a quality judgment - it just keeps the discussion in one place:

The production target serves a static Vite build, so all VITE_* values are frozen at vite build time. Vite loads the committed frontend/.env.production (VITE_API_HOST=https://gptcloud.arc53.com) and inlines it into the bundle.
docker-compose-azure.yaml sets VITE_API_HOST=http://localhost:7091 as a runtime environment: on the nginx container, which has no effect on already-built static files.
Result: the azure stack spins up its own backend/worker/redis/postgres, but the frontend silently talks to the public cloud (gptcloud.arc53.com) instead of the local :7091 backend - a regression from the previous dev-server behavior. (Confirmed by building this branch: localhost:7091 appears nowhere in the bundle; the only arc53.com host present is gptcloud.)

PR #2582, opened 2026-07-06
(feat)workflows: add undo/redo

### LINKED ISSUE #2581: 🚀 Feature: Undo/redo support in the agent workflow builder
### 🔖 Feature description

Add undo/redo to the workflow builder canvas - buttons on the canvas plus Ctrl/Cmd+Z and Ctrl/Cmd+Shift+Z / Ctrl/Cmd+Y shortcuts.

Every canvas action (deleting a node, dropping an edge, moving things, editing node config) is currently irreversible. One misclick - e.g. clicking an edge deletes it instantly means rebuilding by hand. Undo/redo is a baseline expectation for visual editors like this.

A useUndoRedo hook keeping past/future snapshot stacks of {nodes, edges}
Snapshot before each mutating action: node add/delete/drag, edge connect/delete, and config edits — keystroke bursts grouped into a single history entry.
Selection stays out of history: selecting a node creates no entry, and undo/redo doesn't move the current selection.
History clears on workflow load; shortcuts are ignored while typing in inputs so native text undo still works.

### 🔄️ Additional Information

_No response_

ManishMadan2882:
    Closes #2581 
     UX

### DIFF
--- a/frontend/src/agents/workflow/WorkflowBuilder.tsx
+++ b/frontend/src/agents/workflow/WorkflowBuilder.tsx
@@ -582,30 +582,49 @@ function WorkflowBuilderInner() {
     [],
   );
 
+  const deleteNodesAndEdges = useCallback(
+    (nodesToDelete: Node[], edgesToDelete: Edge[]) => {
+      const removableIds = new Set(
+        nodesToDelete.filter((n) => n.type !== 'start').map((n) => n.id),
+      );
+      const edgeIdsToDelete = new Set(edgesToDelete.map((e) => e.id));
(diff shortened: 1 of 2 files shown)


ManishMadan2882:
Pushed a fix unrelated to this PR
- the delete node action made with the backspace or delete key - the issue was that the node was removed from the worflow but the config panel component remained there; core cause of the issue is that the keys passed as  deleteKeyCode={['Backspace', 'Delete']} skipped the handleDeleteNode
- this issue even made is possible to delete the start node which was not the ideal behaviour

pabik (approved):

PR #2601, opened 2026-07-20
Harden Compose database configuration

ferasbusiness666:
## What kind of change does this PR introduce?

Security hardening / configuration fix.

## Why was this change needed?

The default Docker Compose configuration exposed Redis and PostgreSQL ports on the host and used a predictable database password. This can make accidentally internet-facing deployments unsafe.

## What changed?

- Removed default host exposure for Redis (`6379`) and PostgreSQL (`5432`)
- Replaced the fixed PostgreSQL password with a required environment variable
- Updated Linux and Windows setup scripts to generate a secure random `POSTGRES_PASSWORD`

## Validation

- Confirmed both Compose configurations no longer expose Redis or PostgreSQL ports
- Confirmed the hard-coded `docsgpt` database password was removed
- Confirmed both setup scripts generate the password before Docker starts

### DIFF
--- a/deployment/docker-compose-hub.yaml
+++ b/deployment/docker-compose-hub.yaml
@@ -22,7 +22,7 @@ services:
       - CELERY_BROKER_URL=redis://redis:6379/0
       - CELERY_RESULT_BACKEND=redis://redis:6379/1
       - CACHE_REDIS_URL=redis://redis:6379/2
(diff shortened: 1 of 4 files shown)


ferasbusiness666:
hey
what should i do about vercel deployment please

PR #2651, opened 2026-08-10
Fix RandomNumberGenerator.Fill incompatibility with Windows PowerShel…

### LINKED ISSUE #2639: `setup.ps1` fails if you don't have a newer version of PowerShell
I had PowerShell 5.1 on my Windows, which is the default AFAIK on Windows 11 (at least currently):
```
PS C:\Users\USER> $PSVersionTable.PSVersion

Major  Minor  Build  Revision
-----  -----  -----  --------
5      1      26100  8875
```
When I tried to run the PowerShell setup script described in the README, I received the following error:
```
Configuring for Cloud API Provider: Anthropic (Claude)...
.env file configured for Anthropic (Claude).
Method invocation failed because [System.Security.Cryptography.RandomNumberGenerator] does not contain a method named 'Fill'.
At C:\Workspace\DocsGPT\setup.ps1:550 char:9
+         [System.Security.Cryptography.RandomNumberGenerator]::Fill($b ...
+         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
    + CategoryInfo          : InvalidOperation: (:) [], ParentContainsErrorRecordException
    + FullyQualifiedErrorId : MethodNotFound
```

I was able to resolve the issue by installing PowerShell 7, but perhaps the script should be enhanced to either work on PowerShell 5.1 or to check the PowerShell version and give a better error message that a newer PowerShell version is required.

Acestar21:
Fixes #2639

### What kind of change does this PR introduce?
Bug fix.

### Why was this change needed?
`setup.ps1` crashed on Windows PowerShell 5.1 (the default on Windows 11) with:

Method invocation failed because [System.Security.Cryptography.RandomNumberGenerator] does not contain a method named 'Fill'.

`RandomNumberGenerator.Fill()` was introduced in .NET 6 and is only available under PowerShell 7+ (.NET Core/.NET 6+). Windows PowerShell 5.1 runs on .NET Framework, which doesn't have this method, so the script failed immediately when generating the JWT secret key and internal key.

### Fix
Replaced all 3 occurrences of `[System.Security.Cryptography.RandomNumberGenerator]::Fill($bytes)` with the version-compatible pattern:
```powershell
$rng = [System.Security.Cryptography.RandomNumberGenerator]::Create()
$rng.GetBytes($bytes)
$rng.Dispose()
```
This produces identical cryptographically secure random bytes and works on both PowerShell 5.1 and 7+, so no PowerShell version restriction is needed.

### Testing
Tested on Windows PowerShell 5.1.26100.8875:
- **Before fix:** running `setup.ps1` → option 1 crashed at `Ensure-InternalKey` with the error below (screenshot/recording attached).
- 
Screenshot 2026-08-10 181413

https://github.com/user-attachments/assets/27729e30-53cd-4cda-94db-8a8e2045422e

- **After fix:** `setup.ps1` runs cleanly through the full setup flow (Public API Endpoint, Advanced Settings including Simple JWT auth) with no errors (screenshot/recording attached).

https://github.com/user-attachments/assets/e3a8bbd8-795e-4a2b-bf1c-81873dfa9716

### Other information
No other files were changed. This is a minimal, targeted fix scoped to the 3 affected call sites in `setup.ps1`.

### DIFF
--- a/setup.ps1
+++ b/setup.ps1
@@ -434,7 +434,9 @@ function Configure-Auth {
             $jwt_key = Read-Host "Enter JWT secret key (leave empty to auto-generate)"
             if ([string]::IsNullOrEmpty($jwt_key)) {
                 $bytes = New-Object byte[] 32
-                [System.Security.Cryptography.RandomNumberGenerator]::Fill($bytes)
+                $rng = [System.Security.Cryptography.RandomNumberGenerator]::Create()
+                $rng.GetBytes($bytes)
+                $rng.Dispose()
                 $jwt_key = [System.BitConverter]::ToString($bytes).Replace("-", "").ToLower()
                 Write-ColorText "Auto-generated JWT secret key." -ForegroundColor "Yellow"
             }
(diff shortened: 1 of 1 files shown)


ManishMadan2882 (approved):
Tested on Windows PowerShell 5.1, confirmed the crash from #2639 is gone and setup completes as described.

Approving, good catch!

PR #2653, opened 2026-08-10
fix(react-widget): resolve npm audit advisories in shipped dependency tree

ManishMadan2882:
No changes to package.json, this is a lockfile update.

extensions/react-widget ships its production dependency tree to all downstream consumers. That tree had a few security advisories, so this bumps transitive deps to patched versions.

ensures there are no security advisories on running `npm run audit` and `npm run audit --omit-dev`

Other information:
Bundle size increased slightly (approx 7 kB) due to markdown-it update.

### DIFF
--- a/extensions/react-widget/package-lock.json
+++ b/extensions/react-widget/package-lock.json
@@ -67,29 +67,29 @@
       }
     },
     "node_modules/@babel/compat-data": {
-      "version": "7.29.0",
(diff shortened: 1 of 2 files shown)


ManishMadan2882:
remove unused snarkdown dep.

dartpain (approved):

PR #2704, opened 2026-08-30
fix(react-widget): integrate with current contract

ManishMadan2882:
    Bug fix, with some UX improvements. Updates the react-widget to work with the current /stream contract.

     The widget handled only answer, id, source and error. Every other event /stream emits - thought, notice, workflow_step, tool_calls, guardrail, message_id - fell into a catch-all that appended an empty string to the answer, so reasoning, progress and tool activity never rendered and guardrail retractions were ignored. Its SSE reader also split each network chunk on \n with no carry-over buffer and decoded without {stream: true}, so any frame or multi-byte character split across a chunk boundary failed to parse (41 failures vs 0 on a mock with unaligned chunks). It also sent an unread model field and a full history blob the backend discards after the first turn.

That names the four explicitly and puts the unhandled list in its own clause, so there's no way to read it backwards.
- **Other information**:
Widget only, no new dependencies. Embed bundle 400 to 407 kB.

### DIFF
--- a/application/api/user/conversations/routes.py
+++ b/application/api/user/conversations/routes.py
@@ -11,6 +11,7 @@ from application.api.answer.services.conversation_service import (
     TERMINATED_RESPONSE_PLACEHOLDER,
 )
 from application.storage.db.base_repository import looks_like_uuid, row_to_dict
+from application.storage.db.repositories.agents import AgentsRepository
 from application.storage.db.repositories.attachments import AttachmentsRepository
 from application.storage.db.repositories.conversations import ConversationsRepository
 from application.storage.db.repositories.message_events import MessageEventsRepository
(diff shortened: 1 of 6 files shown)


ManishMadan2882:
/api/feedback gated only on request.decoded_token, which is populated solely from a Bearer JWT. Embedded widgets authenticate with an agent api_key and carry no JWT, so every widget feedback request returned 401 on any auth-required deployment  including the widget's default host. The schema declared api_key as an optional field but the handler never read it.

   Inside the existing if not decoded_token block, the key is now resolved to its owner via AgentsRepository.find_by_key, the same primitive /stream and other public endpoints use

dartpain (approved):

PR #2732, opened 2026-09-07
Rename the application package to docsgpt

dartpain:
## Summary

Renames the backend import package from `application` to `docsgpt`. This is the prerequisite for publishing the server to PyPI as `docsgpt` (`pip install docsgpt[docling]`, `uvx docsgpt`): a wheel's import name is its directory name, and `application` is far too generic to install into anyone's site-packages.

The change is mechanical and complete: `git mv application docsgpt`, every import and dotted string rewritten (734 Python files), and every path reference in compose files, Dockerfile, workflows, docs, setup scripts, devcontainer, k8s manifests, vscode config, pytest and coverage config. Nothing else changes behaviour.

## Compatibility for one release

- **Import alias.** A thin top-level `application` package remains. Its meta-path finder resolves `application.x.y` to the already-imported `docsgpt.x.y` object, so there is exactly one settings object, one Celery app and one Flask app whichever name a process uses. `celery -A application.app.celery` and `uvicorn application.asgi:asgi_app` keep working and print a `FutureWarning` pointing at the new names. The image ships the alias.
- **Celery task names.** Tasks are now `docsgpt.api.user.tasks.*` etc. Messages queued by the previous release carry the old names, so the worker registers every `application.*` name as an alias on start-up (`celeryd_after_setup`). The redbeat key prefix is bumped to `redbeat:docsgpt:v2:`, so periodic-task entries the previous release wrote to Redis are left unread instead of firing the old names alongside the new ones. `task_routes` and the no-reclaim list use the new names. The `task_dedup` table's `task_name` column is informational only (reconciliation never re-dispatches by it).
- Both the alias package and the task-name aliases are scheduled for removal in the release after next.

## Docker build context

The backend image now builds from the repository root (`docker build -f docsgpt/Dockerfile .`) so it can carry the alias package next to `docsgpt`, and so later PRs can build from `pyproject.toml`/`uv.lock`. A root `.dockerignore` allow-lists `docsgpt/` and `application/` and re-excludes caches, local data, `.env*`, the sample index files and the Dockerfile itself. Compose files use `context: ..` with `dockerfile: docsgpt/Dockerfile`; the three image workflows use `context: .`. Requirements files and the Dockerfile stay under `docsgpt/` for now.

## Verification

- `ruff check .`; full `pytest` (all green, plus `tests/test_package_rename.py` covering the alias, the once-per-process warning, a missing submodule still raising, every task answering to its old name, idempotent registration, and the new route/prefix values)
- Image built from the root context: 0.89 GB, `application.*` and `docsgpt.*` are the same objects inside it, 25 legacy task names registered, `verify_offline` passes under `--network none`, Dockerfile and sample index files absent from the image
- All compose files validate; workflow YAML parses; docs site builds; requirements exports regenerated from the lock

## Follow-ups (separate PRs, see plan)

1. Packaging: real package, console script, package data, path independence, dependency ranges, `pypi-publish.yml` with trusted publishing.
2. `uvx docsgpt`: frontend bundled in the wheel, `docsgpt up`.

No screenshot applies; this PR has no user-visible UI change.


dartpain:
Addressed the six blocking comments in adb69635:

- **CI installs no backend deps** — `pytest.yml` now does `cd docsgpt` before `pip install -r requirements.txt`.
- **`.dockerignore` re-admits `application/`** — only `!application/__init__.py` is re-admitted and the Dockerfile copies just that file. Rebuilt with a stale `application/inputs` and `application/.env` on the host: `/app/application` in the image holds only `__init__.py`, and `python -m application.scripts.verify_offline` still passes under `--network none`.
- **Compose bind mounts silently move** — host paths stay at `../application/{indexes,inputs,vectors}` in all three compose files (container paths unchanged), with a comment and a line in the upgrade notes. The move comes with the packaging PR, together with an `mv` note.
- **Alias loader breaks `python -m application.<x>`** — `exec_module` restores the real `docsgpt` spec on the shared module object, and `get_code`/`get_source`/`get_filename` delegate to the target loader. Tests cover `importlib.reload` keeping the `docsgpt` name and `runpy.run_module("application.vectorstore.model_registry")`.
- **Celery alias registration flips log names** — each legacy `application.*` name is registered as its own task object (a thin subclass with `name=<legacy>`) via `app.register_task`, so the canonical object keeps its tracer. Test asserts the alias is a distinct instance of the same class and that registration is idempotent.
- **Redbeat prefix bump** — reverted to `redbeat:docsgpt:`; the three `schedule_syncs` entries are named `schedule-syncs-{daily,weekly,monthly}`. No cleanup pass is needed: redbeat tracks static entries in `<prefix>:statics` and deletes the ones that vanish from `beat_schedule` on start-up, and rewrites the task path of named entries in place. Verified against redbeat 2.4.2 with a seeded Redis (old auto-named keys + a named `reconciliation` entry pointing at `application.…`): after `RedBeatScheduler(app)` the old keys are gone from both the hash space and the schedule zset, and the named entries point at `docsgpt.…`. A test now asserts every periodic entry carries a stable name.

Full backend suite: 9986 passed, 383 skipped. Docs build passes.

dartpain:
Second pass, all eight fixed in the latest commit:

- **Sync timers reset on upgrade** — documented in the upgrade note (new entry keys start from `last_run_at = now`; redbeat only preserves meta on existing keys).
- **Reclaim skip vs legacy name** — `_reclaim_memory_after_task` normalises the `application.` prefix before the `_NO_RECLAIM_TASKS` lookup; test added.
- **Azure compose mounts** — now `/app/{indexes,inputs,vectors}`, matching the other compose files and where the process writes. This was already wrong on `main` (`/app/application/...`), so the host directories were never populated; the fix is safe.
- **Verify workflow paths** — `docsgpt/requirements*.txt` replaces the dead `application/requirements*.txt` entry.
- **Dependabot** — pip directory is `/docsgpt`. Separate question for later: Dependabot editing exported requirements files will trip the stale-export check, so switching that entry to the `uv` ecosystem at `/` is probably the right end state.
- **`install_hint()`** — path and docstring say `docsgpt/requirements-<extra>.txt`; the test asserts the full path now.
- **`.gitignore`** — `application/vectors/` added back beside `docsgpt/vectors/`.
- **QA script** — quiets the `docsgpt` logger tree instead of the stale root plus redundant children.

PR #2743, opened 2026-09-07
feat: publish the backend to PyPI as docsgpt

dartpain:
## Summary

The backend is now an installable package: `pip install docsgpt` (extras `docsgpt[docling]`, `docsgpt[milvus]`) with a `docsgpt` command. Second step of the PyPI work, after the package rename in #2732. The frontend build inside the wheel and `docsgpt up` are the next step, not this PR.

- **Build.** hatchling backend, version read from `docsgpt/version.py` (the release workflow reads the same file). The wheel is the `docsgpt` package plus the data it reads at runtime: prompts, model catalogs, seed config, `alembic.ini` and the migrations. It excludes the Dockerfile, the exported requirements files, the sample index and local runtime data; the one-release `application` import alias stays checkout-only. `[tool.uv] package = false` is gone, so `uv sync` installs the package editable and puts `docsgpt` on PATH.
- **`docsgpt` command** (`docsgpt/cli.py`). `api` runs gunicorn with the bounded-drain uvicorn worker and the same flags as the image CMD (`--reload` switches to uvicorn for development). `worker` runs the Celery worker with the scheduler embedded (`--no-beat`, `-Q`, `--concurrency`, `--pool`; solo pool by default on macOS). `migrate` creates the database if needed and migrates to head. `prefetch-models`, `verify-offline` and `reembed` hand their arguments to the existing scripts untouched. `docsgpt --help` does not import the app.
- **Path independence** (`docsgpt/core/paths.py`). Runtime data (`.env`, `inputs/`, `indexes/`) lives in a data home: `DOCSGPT_HOME`, else the checkout, else the current directory; `DOCSGPT_ENV_FILE` overrides the env file. Settings, the dotenv load in `app.py`, `LocalStorage` and the internal upload route use it instead of "three directories above this file", which resolves to `site-packages` once installed. A checkout (detected by its `pyproject.toml`) and the Docker image (no `pyproject.toml`, runs from `/app`) behave exactly as before. Two dead copies of that computation in `worker.py` and `api/user/base.py` are removed.
- **Dependencies.** `[project] dependencies` are compatible ranges (floor = the locked version, ceiling = next major, next minor below 1.0) so the package installs next to other packages. `uv.lock` resolves to the same versions and the exported `requirements*.txt` are unchanged, so Docker, CI and checkouts keep the exact pins.
- **CI.** `package-build.yml` builds the sdist and wheel on PRs that touch the package, checks the wheel contents, installs it into a clean venv and runs the command. `pypi-publish.yml` publishes on `release: published` through trusted publishing (environment `pypi`); a manual run publishes to TestPyPI (environment `testpypi`) or PyPI.
- **Docs.** New page Deploying → Install with pip; a note in AGENTS.md.

## Before the first publish

- Create the `pypi` environment on this repo (add required reviewers if wanted). The pending publisher on PyPI already names `pypi-publish.yml` and the `pypi` environment.
- Optional dry run: add a pending publisher on test.pypi.org for the same workflow with environment `testpypi`, create that environment, and run the workflow manually (default target is TestPyPI).
- The first real publish happens with the next version bump: `backend-release.yml` creates the release, which now triggers the Docker build and the PyPI publish.

## Verification

- Full backend suite, ruff, docs build.
- `uv build`, then the wheel installed into a clean venv: `docsgpt --version`, `--help`, `api --help`. From another directory the installed package resolves no checkout, the data home is the current directory, `.env` is read from there and from `DOCSGPT_HOME`, and `alembic.ini` and the prompt fragments resolve inside `site-packages`.
- `uv lock --check` and the requirements export diff are clean (the same check CI runs).
- zizmor reports nothing on both workflows.

## Summary by CodeRabbit

- **New Features**
  - Added an installable DocsGPT Python package with configurable runtime and environment-file locations.
  - Added the `docsgpt` CLI for serving the API, running workers, applying migrations, and maintenance tasks.
  - API serving defaults to localhost, with an option to listen on all interfaces.
  - Local databases and storage now use the configured runtime data directory.
  - Added command-line help and model selection for model prefetching and offline verification.

- **Documentation**
  - Added pip-install deployment, configuration, service command, upgrade, and runtime-data migration guidance.

- **Chores**
  - Added automated package build validation and publishing workflows.
  - Added safeguards to exclude development and runtime data from distributions.


dartpain:
Review pass, all in the latest commit:

- **CPU torch flow** — the docs and the `pyproject.toml` comment now describe the two-step install (`pip install --index-url https://download.pytorch.org/whl/cpu torch torchvision`, then `docsgpt[docling]`; pip keeps the torch it already has), plus a `pipx inject` equivalent. `--extra-index-url` only yielded the CPU build while that index keeps pace with PyPI.
- **Milvus / LanceDB defaults** — `MILVUS_URI` and `LANCEDB_PATH` derive from `home_dir()` via `default_factory`, so they follow `DOCSGPT_HOME` like the faiss indexes and uploads. Tests cover the resolved defaults under `DOCSGPT_HOME` and that env overrides still win. A checkout run from its root and the Docker image resolve to the same paths as before.
- **Default bind address** — `docsgpt api` binds `127.0.0.1` by default; `--host 0.0.0.0` exposes it. Documented, and the Bandit B104 alert goes away with it.
- **`/dev/shm`** — the two Bandit B108 alerts are dismissed as false positives: gunicorn creates its heartbeat file there with `mkstemp`, and the image passes the same flag.
- **AGENTS.md wording** — `DOCSGPT_HOME` moves the data home; `DOCSGPT_ENV_FILE` selects only the `.env` file.
- **Docs example** — the Postgres URI uses a `<password>` placeholder.

Full suite, ruff and the docs build pass.

dartpain:
Both addressed in the latest commit:

- **pipx** — the docs now use `pipx runpip docsgpt install --force-reinstall --no-deps --index-url https://download.pytorch.org/whl/cpu torch torchvision`. `pipx inject`, with or without `--force`, hands the packages to pip, which reports torch already satisfied and keeps the CUDA build; `--force-reinstall` is what replaces it, and `--no-deps` keeps pip from reinstalling torch's dependencies from the PyTorch index, which carries older copies of them.
- **Local vector-store paths** — the upgrade notes now say that the embedded Milvus and LanceDB default paths resolve under the checkout (or `DOCSGPT_HOME`) instead of the start directory, name the old location (`<start directory>/milvus_local.db`, `<start directory>/data/lancedb`) and the settings to pin (`MILVUS_URI`, `LANCEDB_PATH`), and note that faiss indexes and uploads were already checkout-rooted. No fallback code: it only affects those two stores at their default path when the process is started outside the checkout root.

Docs build passes.

dartpain:
Follow-up review items, all in the latest commit:

- **Docker data home** — `ENV DOCSGPT_HOME=/app` in the runtime stage; the image ships no checkout, so the home no longer depends on `WORKDIR`.
- **API/worker data home** — `api`, `worker`, `beat` and `migrate` print `data home … (env file …)` on start-up; the docs say to run both from one directory or set `DOCSGPT_HOME`.
- **Queues** — `-Q` is passed only when given; a bare worker consumes every configured queue, so renamed `EMBEDDINGS_QUEUE` / `DOCUMENT_PARSE_QUEUE` values are honoured.
- **Script `--help`** — `prefetch_models` and `verify_offline` have argparse now (`models` positional), so `--help` works through `python -m` and through the CLI.
- **Worker exit code** — the worker runs through `celery.start()` and returns its code; click usage errors print usage and exit 2.
- **Windows** — solo pool and no embedded scheduler (celery rejects `-B` there), with a hint to run the new `docsgpt beat` command separately.
- **`DOCSGPT_ENV_FILE`** — a path that is not a file raises at start-up instead of booting with defaults.

Not changed: the gunicorn SIGUSR2 re-exec path under `docsgpt api` (only zero-downtime binary upgrades via USR2 are affected; the image runs the real gunicorn script and is unaffected), and the release-event trigger, which is a separate fix for both the Docker and PyPI publish workflows.

Full suite, ruff, docs build and the Dockerfile lint pass.

PR #2764, opened 2026-09-12
fix(parser): keep markdown sections that follow an empty heading

abo3losh1:
- **What kind of change does this PR introduce?** Bug fix (backend, markdown parser).

`MarkdownParser.markdown_to_tups()` drops a heading whenever the heading before it has no body text. The guard reads:

```python
if current_header is not None:
    if current_text == "" or None:
        continue
```

`continue` skips the rest of the loop body, so `current_header` is never reassigned to the new heading. Everything under it gets folded into the previous section.

That hits the most common markdown layout there is, a title followed immediately by a subheading:

```markdown
# Project Title
## Installation
Run pip install docsgpt.
```

Before: `[("Project Title", "Installation Run pip install docsgpt.")]` — the `Installation` heading is gone, and its text is indexed under `Project Title`.
After: both sections are kept, each with its own text.

Since headings are what the retriever matches on, a lost heading means those chunks are much harder to retrieve for the question they actually answer.

The same function has a second, smaller bug in the heading-less branch. It joins lines with `re.sub("\n", "", value)`, which glues the last word of one line to the first word of the next:

```
Alpha beta
gamma delta
```

became `Alpha betagamma delta`. That creates junk tokens in the embedding for any document with no headings at all. Joining with a space fixes it, and matches what the chunked branch already does.

Note that `current_text == "" or None` is also always just `current_text == ""` — `or None` is dead. Rewriting the guard as `if current_header is not None and current_text != "":` keeps the intended skip-empty-section behaviour without swallowing the heading.

- **Other information**:

Existing test `TestMarkdownParserEmptyHeader::test_empty_text_header_continues` covered this input but only asserted `len(tups) >= 1`, so it passed while the heading was being dropped. It now asserts that `Header 2` survives with its content.

**Validation**

Three tests fail on `main` and pass with the fix:

```
$ python -m pytest tests/parser/file/test_markdown_parser.py tests/test_remaining_coverage.py::TestMarkdownParserEmptyHeader -q

# on main
FAILED tests/parser/file/test_markdown_parser.py::test_markdown_consecutive_headers_are_not_dropped
    assert 'Installation' in ['Project Title']
FAILED tests/parser/file/test_markdown_parser.py::test_markdown_headerless_content_keeps_word_boundaries
    assert 'betagamma' not in 'Alpha betagamma delta'
FAILED tests/test_remaining_coverage.py::TestMarkdownParserEmptyHeader::test_empty_text_header_continues
    assert 'Header 2' in ['Header 1']
3 failed, 4 passed

# with this branch
7 passed
```

`ruff check` (0.14.10) clean on the three touched files. No config, dependency or API changes; frontend untouched, so there is nothing UI-visible to screenshot.

## Summary by CodeRabbit

- **Bug Fixes**
  - Markdown parsing now preserves sections following consecutive headers.
  - Headerless Markdown content retains spaces between words instead of concatenating them.
  - Empty headers are excluded while valid headers and their content remain available.

- **Tests**
  - Added coverage for consecutive headers, headerless content, and empty-header handling.


abo3losh1:
Gentle ping, this has been open about ten days without a review. Still mergeable against `main`, CodeRabbit green, and the only red checks are the two Vercel deploys that need team authorization for a fork.

Short version: `markdown_to_tups()` dropped a heading whenever the heading before it had no body text, because the guard `continue`d before reassigning `current_header`. A README that opens with `# Title` immediately followed by `## Installation` lost the Installation section entirely, and its text was indexed under the title. Since headings are what the retriever matches on, those chunks became much harder to retrieve. The same function also joined heading-less lines with `re.sub("\n", "")`, gluing the last word of one line to the first of the next.

Three tests fail on `main` and pass on the branch, including an existing test that asserted only `len(tups) >= 1` and so passed while the heading was being dropped.

There is also #2790 open on the same repo, a separate fix for `str_replace` in the memory and notes tools passing an unescaped replacement to `re.sub`, which corrupts Windows paths and raises an uncaught `re.PatternError` on a group reference. Both are small if you are opening the repo anyway. Happy to rebase either.

Issue #2766, opened 2026-09-13
🐛 Bug Report: OAuth connector session token theft via postMessage wildcard origin


geo-chen:
### 📜 Description

reported on 15 June 2026 via https://github.com/arc53/DocsGPT/security/advisories/GHSA-949x-3mqg-5xfr

### Summary
The OAuth connector callback status page at `/api/connectors/callback-status` sends the OAuth session token to `window.opener` using `postMessage` with a wildcard target origin (`'*'`). Additionally, the frontend message handler in `ConnectorAuth.tsx` accepts messages from any origin without validating `event.origin`. A malicious page that acts as `window.opener` during an OAuth flow -- for example, by opening a DocsGPT popup from attacker-controlled context -- can receive the connector `session_token`, which links to a stored OAuth credential (Google Drive access token). The `disconnect` endpoint at `/api/connectors/disconnect` additionally accepts this token without any JWT authentication, allowing session destruction without the victim's credentials.

### Details

**Root cause 1 -- server-side wildcard postMessage:**

`application/api/connector/routes.py`, lines 519--524:

```python
if (status === "success" && window.opener) {
    window.opener.postMessage({
        type: providerType + '_auth_success',
        session_token: sessionToken,
        user_email: userEmail
    }, '*');   # <-- wildcard target origin
```

The `session_token` (a UUIDv4 that maps to an OAuth credential row in `connector_sessions`) is broadcast to any origin that is `window.opener` of the callback page.

**Root cause 2 -- client-side missing origin check:**

`frontend/src/components/ConnectorAuth.tsx`, lines 56--79:

```ts
const handleAuthMessage = (event: MessageEvent) => {
    const successProvider = event.data?.type === `${provider}_auth_success`;
    if (successGeneric || successProvider) {
        onSuccess({ session_token: event.data.session_token, ... });
    }
    // event.origin is never verified
};
window.addEventListener('message', handleAuthMessage);
```

Any page can inject a fabricated `session_token` into the DocsGPT frontend by firing a matching message event.

**Root cause 3 -- unauthenticated disconnect endpoint:**

`application/api/connector/routes.py`, lines 347--365 -- `/api/connectors/disconnect` performs no JWT check, accepts a `session_token`, and deletes the matching row from `connector_sessions` without verifying the caller is the token's owner. Live-confirmed: a `POST /api/connectors/disconnect` call without any Authorization header returns `{"success": true}`.

**OAuth credential exposure chain:**

[...]

PR #2767, opened 2026-09-14
fix(connectors): stop broadcasting the OAuth session token (GHSA-949x-3mqg-5xfr)

L4XB:
Refs #2766 / GHSA-949x-3mqg-5xfr. All three root causes are still present on `main`; this closes each of them.

## 1. Wildcard `postMessage` target origin

`docsgpt/api/connector/routes.py` posted the session token with `'*'`, so any page that arranged to be the popup's opener received a handle to a stored OAuth credential.

It now names one origin, resolved server-side:

- `OIDC_FRONTEND_URL` when set — the setting is already documented as "browser-facing app origin", which is exactly this;
- otherwise `request.host_url`, the single-host deployment.

Only scheme + host are used: `postMessage` never delivers to a target carrying a path, so passing a configured URL through unchanged would have broken the flow silently. An unparseable setting falls back to a real origin rather than back to a wildcard — there is a cell for that, because "fall back to something permissive" is how this class of fix usually regresses.

> ⚠️ **One deployment note.** If your app and API are on different hosts and `OIDC_FRONTEND_URL` is unset, the message will now target the API's origin and the app will not receive it. That is the one configuration that needs a value set after this change. I chose to require it rather than keep a wildcard, but if you would rather derive it from something else (a CORS allowlist, a new setting), say which and I will switch to it.

## 2. Missing `event.origin` check in the frontend

`ConnectorAuth.tsx` accepted a `message` from any origin, so any page could inject a fabricated `session_token`. It now listens only to the origin serving the callback page, resolved from `baseURL` (`VITE_API_HOST`) against the page origin, so a relative or empty host resolves to the same-origin case.

## 3. Unauthenticated `/api/connectors/disconnect`

The endpoint took a session token and no credentials: a leaked token was enough to destroy the session it named. It now authenticates through `request.decoded_token` like every other route in the file, **and** requires the session to belong to the caller — authentication alone would still let any logged-in user disconnect a connector whose token they learned.

An unknown token and another user's token get the same 200, so the endpoint does not become an oracle for which tokens exist.

## Tests

13 cells. Three existing disconnect tests now set `decoded_token`; that is the change, not a workaround for it.

| mutation | killed by |
| --- | --- |
| wildcard target origin again | 4 cells, incl. `the_token_is_never_posted_to_a_wildcard_origin` |
| the origin keeps its path | `it_targets_the_configured_browser_app_origin` |
| an unusable setting falls back to `'*'` | `an_unusable_setting_does_not_widen_back_to_a_wildcard` |
| disconnect stops requiring a token | `rejects_an_unauthenticated_caller` |
| disconnect stops checking ownership | `another_users_token_is_not_destroyed` |
| the frontend resolver returns `'*'` | `never_answers_with_a_wildcard, whatever it is given` |
| the frontend resolver keeps the path | 3 cells |

`test_disconnects_session` is the accept control for the two refusals: the owner can still disconnect, and the row is gone afterwards.

```
$ pytest tests/api/test_connector_routes_happy.py tests/api/test_connector_routes.py
49 passed

$ npx vitest run src/components/ConnectorAuth.originCheck.test.ts
5 passed

$ npx tsc --noEmit        # frontend, clean
$ npx eslint src/components/ConnectorAuth.tsx
3 warnings — identical on main with this branch stashed
```

**What I could not verify:** the end-to-end OAuth round trip. It needs real Google Drive credentials and a browser, so the popup → opener handoff is exercised at the unit boundary (the rendered page's target origin, and the resolver the listener compares against) rather than by completing a real flow. That is the part worth a maintainer's eye before release.

## Summary by CodeRabbit

* **Security Enhancements**
  * OAuth callback messages are now sent only to the configured or verified browser origin, with safe fallback handling for invalid origins.
  * Connector authentication ignores messages from unexpected origins.
  * Disconnect requests require authentication and can only remove the requesting user’s connector session.

* **Tests**
  * Added coverage for origin validation, secure callback messaging, authentication requirements, malformed origins, and connector ownership checks.


L4XB:
Valid, and worse than a cosmetic guard — the exception escapes the handler, so a misconfigured `OIDC_FRONTEND_URL` turns the callback page itself into a 500 and the popup never delivers the token at all. That is a harder failure than the one the setting exists to fix.

Measured, both halves of the finding:

```
urlparse("https://[broken")                 -> ValueError: Invalid IPv6 URL
urlparse("https://app.example.com:notaport") -> parses fine
  .port                                      -> ValueError: Port could not be cast to integer value
```

So the second shape never raised in `urlparse` — it would have been rendered as a `postMessage` target with a port no browser accepts. Both now fall back to the request's own origin, which is the fallback an unparseable setting already took. Still never `"*"`.

Test: `test_a_setting_that_cannot_be_parsed_falls_back_instead_of_500ing`, parametrised over both shapes.

| mutation | result |
| --- | --- |
| remove the `try/except` entirely | both parameters FAIL |
| catch `urlparse` but not `.port` | the `:notaport` parameter FAILS, `[broken` passes |

`pytest tests/api/test_connector_routes_happy.py` → 39 passed. `ruff check` clean.

(The two Vercel checks are the preview deploys waiting on an Arc53 team member to authorize the build for a fork — nothing I can act on from this side.)

L4XB:
Closing this: `main` already carries the fix, and a more complete one than this PR.

`41b3afed` ("fix: stop leaking connector OAuth session tokens to other origins") and `08e8de73` landed on 2026-09-14 and cover both halves, plus one this PR did not:

- **Sender.** `connector_allowed_origins()` builds an allowlist from the request host, `CONNECTOR_REDIRECT_BASE_URI`, `OIDC_FRONTEND_URL` and `CONNECTOR_ALLOWED_ORIGINS`, and the popup posts to each of them by name. That is better than what I sent, which named a single origin and would have broken the split-origin and loopback dev setups the allowlist handles deliberately. `_origin_of` also returns `None` on an unparseable value rather than raising, which is the follow-up commit on this branch made redundant.
- **Receiver.** `isTrustedConnectorMessage(event, authWindowRef.current, callbackOriginRef.current)` checks the source window and the origin, which is at least as strong as the check here.
- **Ownership.** `delete_by_session_token(session_token, user_id)` now filters on the user in SQL. This branch did that with a read-then-compare in Python, which the SQL form supersedes.

Coverage is upstream too, in `tests/api/test_connector_oauth_security.py` and `frontend/src/utils/connectorAuthUtils.test.ts`, so there is nothing here worth rebasing onto it either.

The branch went `CONFLICTING` against all of that, which is how I noticed. Merging it would mean landing a second, narrower implementation of a fix that is already in, so it is going rather than being resolved. Thanks for the quick turnaround on the report.

PR #2773, opened 2026-09-14
fix(parser): strip standard RST directives in RstParser

### LINKED ISSUE #2770: 🐛 Bug Report: RstParser does not remove standard reStructuredText directives
### 📜 Description

`RstParser.remove_directives()` does not recognize standard reStructuredText directive syntax.
The current regex expects a backtick before the directive and does not handle the required space between `..` and the directive name:
```python
pattern = r"`\.\.([^:]+)::"
```
Therefore, valid directives such as `.. note::` remain in the parsed content.

### 👟 Reproduction steps

1. Create an instance of `RstParser`.
2. Pass the following content to `remove_directives()`:
   `Text before\n.. note:: Important information\nText after`
3. Examine the returned string.
4. Observe that the valid `.. note::` directive marker remains unchanged.

### 👍 Expected behavior

The parser should remove the standard reStructuredText directive marker while preserving useful inline content.
Expected result:
`Text before\nImportant information\nText after`

The method returns the directive unchanged:
`Text before\n.. note:: Important information\nText after`
This is a parser-level bug, so a UI screenshot is not available. A regression test can demonstrate the failure.

### 💻 Operating system

Windows

_No response_

Local dev server

_No response_

The existing `test_remove_directives` test uses the nonstandard form `` `..note::` ``. It therefore does not verify valid reStructuredText syntax such as `.. note::`.
Relevant files:
- `docsgpt/parser/file/rst_parser.py`
- `tests/parser/file/test_rst_parser.py`

I am willing to add regression coverage and submit a focused PR.

### 📖 Relevant log output

```shell

```

### 🧑‍⚖️ Code of Conduct

lorenzozanee:
- **What kind of change does this PR introduce?** Bug fix in the RST parser.

- **Why was this change needed?** `RstParser.remove_directives()` did not recognize standard reStructuredText directive syntax. Its pattern required a backtick before the marker and no space after `..`, so valid directives such as `.. note::` survived parsing while only the nonstandard `` `..note::` `` form was stripped. Fixes #2770.

  The pattern now matches the standard `.. name::` marker (with the required space) while keeping the legacy backtick form working, and preserves the directive's inline content: `.. note:: Important information` yields `Important information`.

- **Other information**: Added `test_remove_directives_standard_syntax` covering the issue's exact repro plus `warning` / `code-block` variants and the legacy backtick form. All 18 tests in `tests/parser/file/test_rst_parser.py` pass; `ruff check` is clean on both touched files.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of reStructuredText directives, including standard syntax and directives with varied spacing.
  - Preserved compatibility with the legacy backtick format.
- **Tests**
  - Added coverage for note, warning, and code-block directives.


L4XB:
Heads up on the pattern here, because I think this one deletes prose it should leave alone.

`r"(^|\s)`?\.\.\s*[^:\n]+::[ \t]*"` has `[^:\n]+` greedy and unanchored, and `\.\.` will happily match the first two dots of an ordinary `...` ellipsis. `::` at end of line is RST's literal-block marker, so an ellipsis followed later by `::` is not an exotic input, it is ordinary documentation:

```
input : 'Install it ... then run the following::'
this PR: 'Install it '
#2772   : 'Install it ... then run the following::'   (unchanged)
```

Everything between the ellipsis and the marker is gone.

`\.\.\s*` also lets `\s` match a newline, so it can swallow across lines:

```
input : 'end of paragraph ..\n\nNext line has a marker:: here'
this PR: 'end of paragraph here'
#2772   : unchanged
```

Smaller, but also worth a look: the leading indentation is left behind, because the pattern consumes the whitespace as its `(^|\s)` group rather than anchoring to the start of the line.

```
input : 'Body\n    .. warning:: Careful'
this PR: 'Body\n    Careful'
#2772   : 'Body\nCareful'
```

All measured on `main` at `d121eec2` with each patch applied in turn, against the parser directly.

Anchoring per line rather than on `\s` closes all three, which is what #2772 does with `^[ \t]*\.\.[ \t]+[\w-]+::[ \t]*` under `MULTILINE`: requiring `[\w-]+` for the directive name means an ellipsis cannot stand in for one, and `^[ \t]*` takes the indentation with it. I ran both patterns over the same battery, including the must-not-strip cases, and that one came out clean on all of them.

Not trying to pick a winner between the two PRs, just flagging that the current tests here would not catch any of this.

PR #2791, opened 2026-09-16
fix(markdown): keep the word boundary a styled space carries

L4XB:
- **What kind of change does this PR introduce?** Bug fix.

A word boundary disappears from every HTML document DocsGPT ingests when the space between
two words sits inside an inline element.

`markdownify` runs an inline element's text through `chomp()`, which lifts the surrounding
whitespace out of the text and then returns `''` once nothing is left. An element holding
**only** whitespace therefore disappears together with its whitespace:

| html | before | after |
|---|---|---|
| `Hello<b> </b>world` | `Helloworld` | `Hello world` |
| `Hello<em> </em>world` | `Helloworld` | `Hello world` |
| `Hello<a href="…"> </a>world` | `Helloworld` | `Hello world` |
| `<b>First</b><b> </b><b>Last</b>` | `**First****Last**` | `**First** **Last**` |

All ten inline tags tested lose it (`a b strong em i s del code sub sup`), measured through
`markdownify(..., **MARKDOWNIFY_OPTIONS)`.

The last row is what makes it ordinary rather than an edge case: an editor that emits one
element per styled run puts the space between two bold words in an element of its own, so any
page that bolds a phrase mid-sentence can produce it. Two words then enter the chunker and the
embedder as one token sequence, and neither word's query matches it well. The same text is
what `read_webpage` hands the model.

- **Other information**:

Three call sites convert HTML to Markdown and `html_parser.py` already documents that they are
meant to produce the same shape ("The crawler's conventions … so file and web ingestion produce
the same Markdown shape"), with the options copied into each:

- `parser/file/html_parser.py` — `soup_to_markdown`
- `parser/remote/crawler_markdown.py`
- `agents/tools/read_webpage.py`

`docsgpt/parser/markdown_conversion.py` is new and now holds `MARKDOWNIFY_OPTIONS` once plus
`WhitespacePreservingConverter`, which wraps the conversion function markdownify resolves for
a tag so a whitespace-only element returns its text unchanged. One hook covers all fourteen
inline tags rather than overriding each of them. All three call sites go through it;
`html_parser` keeps its lazy-import shape, so nothing new is imported at module load.

Two existing tests patched `crawler_markdown.markdownify` and
`read_webpage.markdownify` by name; they now patch `html_to_markdown_text` at the same points
and assert the same things.

`tests/parser/test_markdown_conversion.py` is new, 18 cases: the whitespace-only element across
ten tags, the two-styled-runs case, `&#160;` staying a non-breaking space, five controls with
real content (including a genuinely empty `<b></b>`, which still yields `Helloworld` because
there is no whitespace to keep), and the ATX/backslash conventions plus an override.

```
tests/parser/ + tests/agents/     2918 passed, 48 skipped   (main: 2900 passed, 48 skipped)
ruff check                        All checks passed
```

## Summary by CodeRabbit

- **Improvements**
  - Improved webpage and crawler HTML-to-Markdown conversion.
  - Preserved whitespace in inline content to prevent words from running together.
  - Standardized Markdown formatting, including headings and line breaks.
  - Conversion settings can now be adjusted where needed while retaining consistent defaults.

- **Tests**
  - Added coverage for whitespace preservation, styled content, non-breaking spaces, empty elements, and customizable formatting options.


L4XB:
On the outside-diff note for `tests/agents/tools/test_read_webpage_errors.py`: confirmed and fixed in b2ae49ef4. With the old setup the tool returned `Error fetching URL https://example.com/: '<=' not supported between instances of 'int' and 'MagicMock'`, and `html_to_markdown_text` was never called. The mocked response now answers `200`. The test asserts that the converter was called with `<h1>hi</h1>` and that the result is `Error fetching URL https://example.com/: boom`.

PR #2792, opened 2026-09-16
fix(pptx): read groups and tables, and stop shapes running together

L4XB:
### What

`PPTXParser.parse_file` keeps a shape only when it has a `text` attribute, and
concatenates with no separator:

```python
for shape in slide.shapes:
    if hasattr(shape, "text"):
        slide_text += shape.text
```

Two things fall out of that, both silent.

**1. Groups and tables are dropped.** A group carries no text of its own — its
children do — and the graphic frame a table lives on has no `text` either. So
every shape inside a group and every table cell is missing from the Document.

**2. Shapes run into each other.** With no separator, the last word of one shape
is glued to the first word of the next.

Measured on one slide holding two text boxes, a 2x2 table and a group of two
boxes, the entire extraction was:

```
'FIRST BOXSECOND BOX'
```

| | before | after |
|---|---|---|
| `FIRST BOX` | yes | yes |
| `SECOND BOX` | yes | yes |
| `HEADER A` (table) | **no** | yes |
| `CELL B` (table) | **no** | yes |
| `GROUPED ONE` | **no** | yes |
| `GROUPED TWO` | **no** | yes |
| `BOXSECOND` | **yes** | no |

`BOXSECOND` is a word that appears nowhere in the deck, and it is what gets
embedded and indexed.

### How

A small `_shape_text` returning the blocks of text a shape carries:

- a group recurses (a group can hold a group). It is identified by having its
  own `shapes` collection, which is duck-typed so the existing tests' fake
  `pptx` module keeps working;
- a table renders as a GFM table — the form `tableize` already emits for PDF
  output, so tables read the same however they entered the index;
- anything else falls back to `shape.text` as before.

Blocks are joined with a newline, and each block is stripped rather than only the
finished slide.

### The two existing tests

`test_pptx_parser_concat_true` and `test_pptx_parser_list_mode` asserted the
concatenation, using a fake whose shape texts carried trailing spaces
(`["Hello ", "World"] -> "Hello World"`). That trailing space is what made the
missing separator look correct; a real PowerPoint shape does not end with one.
Their expectations move to the newline form (`"Hello\nWorld\nSlide2"` and
`["A\nB", "C"]`). Nothing else in them changes.

### Tests

Four tests added to `tests/parser/file/test_pptx_parser.py`, building a real deck
in a tmp_path so the fake cannot hide the behaviour again: shapes do not run
together, table cells are extracted as a GFM table, text inside a group is
extracted, and an empty slide is still an empty string.

```
$ pytest tests/parser/file/test_pptx_parser.py -q
8 passed

# same tests, docsgpt/ reverted to main
5 failed, 3 passed
```

`ruff check` clean at the version CI pins (0.16.7).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved PowerPoint content extraction by preserving separate text blocks as newline-separated content.
  * Added support for extracting text from grouped shapes and table cells.
  * Tables are now formatted as Markdown, with cell content safely preserving table structure.
  * Empty slides now produce empty content without stray text.

* **Tests**
  * Expanded coverage for shape, table, grouped-content, and empty-slide extraction scenarios.


L4XB:
Pushed `819768c`, which closes a gap I left in this PR.

Escaping the pipe was not enough. A slide table cell with more than one paragraph carries a **line break**, and `cell.text` hands it over verbatim, so the row ended in the middle of itself:

```
| Step | Notes |
| --- | --- |
| 1 | first line
second line |
```

That is no longer a table — the row is cut in half and `second line |` becomes a paragraph, so everything after it stops being a table too. A multi-paragraph cell is what you get by pressing Enter inside a cell in PowerPoint.

The second gap was in the escape itself: `replace("|", "\\|")` on a cell holding `a\|b` produces `a\\|b`, which is one literal backslash followed by an *unescaped* pipe — the cell's own backslash consumes the escape and the column break comes back.

Both are handled now:

- a line break inside a cell (with surrounding whitespace) collapses to a single space, which is the only thing a GFM row can hold;
- the run of backslashes in front of a pipe is doubled before the pipe is escaped.

```
| Step | Notes |
| --- | --- |
| a\|b | first line second line |
```

New test `test_a_cell_cannot_break_the_table_row` pins both in one table; it fails on the previous commit of this branch and passes now. The PPTX suite is 9 passed, and `ruff check` is clean.

PR #2797, opened 2026-09-16
fix(parser): keep the rows under a rowspan in their own columns

L4XB:
### What this fixes

A GFM table has no way to merge cells downwards, so a cell carrying a `rowspan` is written once and **every row it reaches into comes out one cell short**.

On `main` (2010af3), through `HTMLMarkdownParser`:

```
input:   <table><tr><th>Region</th><th>Product</th><th>Units</th></tr>
           <tr><td rowspan="2">EU</td><td>Cable</td><td>12</td></tr>
           <tr><td>Hub</td><td>7</td></tr>
           <tr><td>US</td><td>Cable</td><td>3</td></tr></table>

output:  | Region | Product | Units |
         | --- | --- | --- |
         | EU | Cable | 12 |
         | Hub | 7 |            <- two cells against a three-column header
         | US | Cable | 3 |
```

Read back, `Hub` sits under **Region** and `7` under **Product**. `Units` is empty for that row and the count is silently attributed to the wrong column, which is worse than losing it — nothing about the output looks wrong, so the chunk goes into the index with the values shifted.

A merged first column is ordinary markup: a region spanning several product rows, a category spanning its items, a date spanning a day's entries. It is also what a Word or Excel table with a vertically merged cell converts to.

markdownify pads a `colspan` (it emits the extra empty cells) but has nothing for `rowspan`, so this reaches every page and file that goes through `soup_to_markdown`.

### The change

`soup_to_markdown` fills the rows a span reaches into with the empty cells it stands for, before markdownify lays the table out.

`_fill_row_spans` walks each table once and builds the grid the way HTML defines it — a cell takes the next column not already taken by a span from above, then claims `colspan` columns across and `rowspan` rows down. Each row that ends up with claimed columns gets an empty `<td>` at each of them.

Two details worth calling out:

- placeholders are inserted in **descending** column order, each directly in front of the first own cell at or after its column, so no index arithmetic is needed and the order comes out right when several land next to each other;
- `_span` clamps to `1..1000`, the same bounds markdownify already applies to `colspan`, so a hostile `rowspan="999999"` cannot blow up the row count.

A table with no `rowspan` is untouched, and `colspan` keeps being handled by markdownify exactly as before.

### Tests

Five new cases in `tests/parser/file/test_html_parser.py`. They read the produced table back the way a reader does and assert on the cells:

- the three-column example above;
- a span in the **last** column, where the placeholder has to be appended rather than inserted;
- a cell spanning in both directions, filling two columns on the next row;
- two guards that must not move: a plain table and a `colspan`-only table.

Measured:

| | result |
|---|---|
| with the change | 28 passed |
| source change stashed, tests kept | **3 failed**, 25 passed |

`tests/parser`: all passing. `ruff check` clean.

> Note: #2796 touches the same file, in `MARKDOWNIFY_OPTIONS`. The two are independent; I will rebase whichever lands later.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved HTML-to-Markdown table conversion for cells spanning multiple rows or columns.
  * Preserved correct column alignment when row and column spans are combined, including spans in the final column.
  * Added support for `rowspan="0"` according to row-group boundaries.
  * Kept tables without row spans unchanged.
  * Improved reliability and performance for very large or complex tables while preventing excessive output.


L4XB:
Pushed 5f01e161 after reviewers of the same padding logic elsewhere (QwenLM/qwen-code#12039, LearningCircuit/local-deep-research#6520) measured two costs that apply here too, both reachable from any HTML file:

| input | 2e59bbbb | 5f01e161 |
| --- | --- | --- |
| one cell `rowspan="1000" colspan="1000"` + 999 rows | 20.5 s | well under a second |
| 30,000 `rowspan="2"` cells, one cell below | 11.5 s | 1.0 s |

- **Unbounded placeholders** — a rowspan now reaches no further than the table's last row, and a table gets placeholders only while they stay within 64 + 8 per real cell; past that it is converted exactly as markdownify converts it today.
- **Quadratic insertion** — `insert_before` scans the siblings once per placeholder. The row is now rebuilt in one pass: children taken out front to back (every `extract()` finds its node at index 0) and put back with the placeholders in place.

Two tests under a 5 s bound (`test_a_span_attribute_does_not_blow_up_the_output`, `test_many_rowspans_are_filled_in_linear_time`) fail on 2e59bbbb and pass here; the existing rowspan cases are unchanged. `tests/parser/`: 1255 passed. `ruff check` clean.

L4XB:
Added one more case to this branch: `rowspan="0"`.

`_span` accepted it as a digit string and clamped it to 1 (`max(1, min(1000, 0))`), so a `rowspan="0"` cell padded nothing and the rows under it came out one cell short — the same misattribution this PR fixes for a numeric rowspan:

```
| Region | Product | Units |
| --- | --- | --- |
| EU | Cable | 12 |
| Hub | 7 |
| Dock | 3 |
```

`rowspan="0"` is the one span HTML states as a range instead of a count: the cell reaches the last row of its row group. Chrome reports `rowSpan === 0` and makes the cell as tall as the whole `tbody`, so rows 2 and 3 above belong in the second column. The layout now spans such a cell to the end of its `thead`/`tbody`/`tfoot` (or of the table, for rows written directly under it), and the next group keeps its own first column:

```
| Region | Product | Units |
| --- | --- | --- |
| EU | Cable | 12 |
|  | Hub | 7 |
|  | Dock | 3 |
```

`colspan="0"` stays one column (HTML5 dropped it; `colSpan === 1`), and a numeric prefix reads as before. The placeholder budget is unchanged and still applies — a `rowspan="0"` asks for no more rows than its group has: `<td rowspan="0" colspan="1000">` over 1000 rows converts in 0.07 s with output 0.89x the input, and 20,001 rows under one `rowspan="0"` in 1.59 s.

Four tests added. Before the change 3 of them fail (3 failed, 31 passed); after, `tests/parser/file/test_html_parser.py` is 34 passed. `ruff check .` clean at the pinned 0.16.7, full suite `10290 passed, 426 skipped`.

